PR #134, opened 2026-02-24
feat: A2A channel, internal channel, AG-UI provider + security hardening

automagik-genie:
## Summary
Split from PR #121 (council review: REJECT — too large). Adds A2A protocol support, internal agent-to-agent routing, and AG-UI provider integration with council-mandated security hardening.

### New Packages
- `channel-a2a`: JSON-RPC + SSE server for A2A protocol clients
- `channel-internal`: Instance-to-instance agent chaining (hop limit: 5)

### New Providers
- `A2AAgentProvider`: Calls external A2A-compatible agents
- `AgUiAgentProvider`: Streams to/from CopilotKit AG-UI agents

### Security Hardening (Council-Mandated)
- **Hop limit fix**: hopCount now increments on each forward (was never incremented — infinite loop vulnerability)
- **A2A input validation**: Zod schema validates all JSON-RPC params (was unsafe cast)
- **A2A feature flag**: `A2A_ENABLED` env var required to activate endpoints (defaults to off)

### Breaking Changes
Drops legacy columns from `agent_routes`:
- `agent_routes.agent_id` (varchar) → replaced by uuid FK
- `agent_routes.agent_provider_id` → removed
- `agent_routes.agent_type` → removed

## Merge Order
**4 of 6** from PR #121 split. Depends on: `feat/agent-entity-model` (PR #130).

This is a stacked PR — includes Group B's changes as base.

## Test Plan
- [ ] `make check` passes (typecheck + lint + dead-code: ✅)
- [ ] 144 unit tests pass across a2a, internal, and core/events packages
- [ ] A2A handler rejects invalid params with -32602
- [ ] A2A endpoints return 503 when A2A_ENABLED is unset
- [ ] Hop limit enforced on internal channel forwarding

PR #133, opened 2026-02-24
feat: agent tasks table + agent state machine (NATS KV + SSE)

automagik-genie:
## Summary
Split from PR #121 (council review: REJECT — too large). Adds persistent agent task tracking and ephemeral agent state machine.

- Agent tasks table (migration 0014) with full lifecycle tracking
- Agent state via NATS KV bucket (ephemeral, high-frequency updates)
- SSE endpoint for real-time agent status streaming
- CRUD routes for agent tasks
- New AGENT NATS stream (stream count 9→10)

## Merge Order
**5 of 6** from PR #121 split.
Depends on: `feat/agent-entity-model` (PR #130) AND `feat/conversations-table` (PR #132).

This is a stacked PR — includes Group B + C changes as base. After those merge, this PR's diff shrinks to just agent tasks + state changes.

## Test Plan
- [ ] `make check` passes
- [ ] Agent tasks CRUD returns 200
- [ ] Agent state SSE endpoint streams updates
- [ ] NATS AGENT stream configured

PR #154, opened 2026-02-27
feat(agent-replay): auto-replay missed messages on reconnect

vasconceloscezar:
## Summary

- Auto-replay missed messages when API reconnects (24h window)
- Instance-level `replayEnabled` flag (default: `true`) and `lastSeenAt` tracking in DB
- `omni replay <instanceId> [--since <timestamp>]` CLI command for manual trigger
- `POST /api/v2/instances/:id/replay` REST endpoint for manual trigger

## How it works

On `instance.connected`, `AgentReplayService.onInstanceConnect()` is called (fire-and-forget):

1. Checks if `replayEnabled` is true for the instance
2. Uses `lastSeenAt` as the window start (skips replay if `lastSeenAt` is null — first connect)
3. Hard-caps the window at 24h ago to avoid unbounded replays
4. Queries all inbound (non-agent, non-deleted) messages for the instance within the window
5. Re-publishes each as a `message.received` event with `source: 'agent-replay'` so the normal agent dispatcher pipeline handles them
6. Updates `lastSeenAt` to now (also updated on `instance.disconnected` for accurate tracking)

The schema adds two new columns to the `instances` table:
- `replay_enabled boolean NOT NULL DEFAULT true` — opt-in/out per instance
- `last_seen_at timestamp` — tracks last clean connect/disconnect for the replay window

## Test plan

- [ ] Start API, send a message to a connected instance — verify it is processed normally
- [ ] Restart API — verify the missed message is replayed on reconnect
- [ ] Set `replayEnabled=false` on an instance — verify no replay happens on reconnect
- [ ] Run `omni replay <instanceId>` manually — verify messages are re-dispatched
- [ ] Run `omni replay <instanceId> --since 2026-02-27T00:00:00Z` — verify only messages after that timestamp are replayed
- [ ] Verify messages older than 24h are NOT replayed (window is capped)
- [ ] Verify agent-sent messages (senderAgentId IS NOT NULL) are NOT replayed

### DIFF
--- a/packages/db/drizzle/0003_real_marvel_zombies.sql
+++ b/packages/db/drizzle/0003_real_marvel_zombies.sql
@@ -1,0 +1,5 @@
+ALTER TABLE "instances" ADD COLUMN "replay_enabled" boolean DEFAULT true NOT NULL;--> statement-breakpoint
+ALTER TABLE "instances" ADD COLUMN "last_seen_at" timestamp;--> statement-breakpoint
+ALTER TABLE "instances" ADD COLUMN "agent_chain_to_instance_id" uuid;--> statement-breakpoint
+ALTER TABLE "instances" ADD COLUMN "chain_mode" varchar(20) DEFAULT 'off' NOT NULL;--> statement-breakpoint
+ALTER TABLE "instances" ADD CONSTRAINT "instances_agent_chain_to_instance_id_instances_id_fk" FOREIGN KEY ("agent_chain_to_instance_id") REFERENCES "public"."instances"("id") ON DELETE set null ON UPDATE no action;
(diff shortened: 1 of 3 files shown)


namastex888:
## Migration Merge Order Requirement

This PR contains migration `0002`. To avoid migration numbering collisions, these PRs must merge in strict sequential order:

1. **PR #102** (`fix/post-streaming-fixes`) -- merge first (migration `0002`)
2. **PR #154** (this PR) -- rebase onto `dev` after #102 merges, regenerate migration to `0003`
3. **PR #133** (`feat/agent-tasks-state`) -- rebase onto `dev` after #154 merges, regenerate migration to `0004`
4. **PR #134** (`feat/a2a-internal-channels`) -- rebase onto `dev` after #133 merges, regenerate migration to `0005`

**After the preceding PR merges:**
```bash
git checkout fix/agent-replay-missed-messages
git rebase origin/dev
cd packages/db && bunx drizzle-kit generate  # regenerates as 0003
git add packages/db/drizzle/ && git commit -m "chore(db): regenerate migration after rebase"
git push --force-with-lease
```

This is by design per the merge-all-open-prs dependency graph (Groups C->D->E->F are strictly sequential).

Issue #178, opened 2026-03-06
Access rules: LID not resolved to phone number before allowlist check


vasconceloscezar:
## Bug Description

When a WhatsApp instance is configured with `accessMode: allowlist` and access rules use `phonePattern`, messages sent by users identified via LID (Linked ID) format are denied even when the user's phone number is in the allowlist.

## Root Cause

In `agent-dispatcher`, the access check runs against the raw `from` value from the message event. When WhatsApp uses LID format (e.g., `7073995747482@lid`), the dispatcher checks the LID directly against `phonePattern` rules — without first resolving the LID to its corresponding phone number.

This means any user whose contact is stored as a LID on the server side will always be denied on allowlist instances, even if their phone is explicitly allowed.

## Steps to Reproduce

1. Create an instance with `accessMode: allowlist`
2. Add an access rule with `phonePattern: "5511968422008"` (phone number)
3. The user's WhatsApp connection sends messages with LID `7073995747482@lid` instead of their phone
4. Observe log: `Access denied — No matching rule, default deny (allowlist)`

## Log Evidence

```json
{"level":"info","module":"message","msg":"Received","from":"7073995747482","chatId":"7073995747482@lid"}
{"level":"info","module":"agent-dispatcher","msg":"Access denied","instanceId":"728d27f2-...","chatId":"7073995747482@lid","from":"7073995747482","reason":"No matching rule, default deny (allowlist)"}
```

## Expected Behavior

Before checking access rules, Omni should resolve the LID to its phone number (via `chatIdMappings` or contact lookup) and check both the LID and the resolved phone against the rules.

## Workaround

Add the user's LID as a `platformUserId` rule alongside their `phonePattern` rule:

```json
POST /api/v2/access/rules
{
  "instanceId": "...",
  "ruleType": "allow",
  "platformUserId": "7073995747482@lid",
  "action": "allow",
  "enabled": true
}
```

Or switch to `accessMode: blocklist` to allow all by default.

## Environment

- Omni v2 (WhatsApp Baileys channel)
- LID format contacts (`@lid` suffix)

PR #193, opened 2026-03-11
fix(access): resolve LID to phone number before allowlist/blocklist check (GH#178)

namastex888:
## Summary

- **Bug**: When WhatsApp LID-first mode is enabled, sender identities are LIDs (e.g., `100000001`) instead of phone numbers. Access rules match phone numbers, so LID-addressed senders were incorrectly denied or unmatched even when a valid phone rule existed.
- **Root cause**: `checkAccessWithFallback` only checked the raw LID ID and `participantAlt` from the message key (only available for group messages), never consulting the in-memory LID mapping cache or `remoteJidAlt` for DMs.
- **Fix**: Two-part change:
  1. `channel-whatsapp/src/handlers/messages.ts` — added `annotateSenderResolvedPhone` which resolves LID → phone using `participantAlt` (groups), `remoteJidAlt` (DMs), or the LID mapping cache, then writes `resolvedSenderPhone` onto the raw message so it flows through `rawPayload`.
  2. `api/src/plugins/agent-dispatcher.ts` — `checkAccessWithFallback` now uses `rawPayload.resolvedSenderPhone` as a third fallback when both primary LID ID and `participantAlt` fail the access check.

Fallback behaviour for **unresolvable LIDs is unchanged** (deny on allowlist, allow on blocklist).

## Test plan

- [x] `allows LID sender when resolvedSenderPhone matches allowlist rule` — verifies the fix path
- [x] `denies LID sender when resolvedSenderPhone is also not in allowlist` — verifies deny still works
- [x] `allows LID group sender via participantAlt when resolvedSenderPhone is absent` — verifies existing group fallback unaffected
- [x] All 2214 existing tests pass (CI full suite)

Closes: #178
NAM-54

### DIFF
--- a/packages/api/src/plugins/__tests__/agent-dispatcher.test.ts
+++ b/packages/api/src/plugins/__tests__/agent-dispatcher.test.ts
@@ -494,6 +494,95 @@ describe('agent-dispatcher', () => {
       expect(services.agentRunner.run).not.toHaveBeenCalled();
     });
 
+    it('allows LID sender when resolvedSenderPhone matches allowlist rule', async () => {
+      const eventBus = createMockEventBus();
+      // Access check: deny LID ID but allow phone number (simulates allowlist with phone rule)
+      const checkAccess = mock(async (_instance: unknown, id: string) => {
+        if (id === '5511999000001') return { allowed: true, reason: 'Phone matched allowlist' };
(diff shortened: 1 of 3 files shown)


namastex888:
@codex review

PR #422, opened 2026-04-16
fix(api,cli): instance PATCH + CLI expose messageSplitDelay* fields

vasconceloscezar:
## Summary

Fixes #373 — the DB schema has \`messageSplitDelayMode/FixedMs/MinMs/MaxMs\` and the dispatcher honors them at runtime, but the \`PATCH /instances/:id\` zod schema doesn't list them and \`omni instances update\` has no flags for them. Configuring split delays required direct DB updates.

## Changes

- **\`packages/api/src/routes/v2/instances.ts\`** — add the 4 fields to \`createInstanceSchema\` (with DB defaults: mode=randomized, fixed=0, min=300, max=1000); extend \`updateInstanceSchema\` to strip defaults on PATCH so an omitted field doesn't silently reset a live instance back to defaults (same pattern as readReceipts/reactionAck).
- **\`packages/cli/src/commands/instances.ts\`** — new \`applySplitDelayFields\` helper mirroring the debounce helper, wired into \`buildInstanceBody\`. Four new flags on both \`create\` and \`update\`:
  - \`--split-delay-mode <disabled|fixed|randomized>\`
  - \`--split-delay-fixed <ms>\`
  - \`--split-delay-min <ms>\`
  - \`--split-delay-max <ms>\`

## Test plan

- [x] \`bun run typecheck\` clean across 19 packages
- [x] Biome lint clean
- [ ] Manual verify: \`omni instances update <id> --split-delay-mode fixed --split-delay-fixed 500\` persists to DB
- [ ] Manual verify: omitting the fields on PATCH does NOT reset live instance back to defaults

Note on test coverage: this is a mechanical schema-expansion on top of columns that already exist in the DB with NOT NULL defaults. Drizzle's typed \`update()\` call ensures type-safety down-stack. No new runtime behavior — the dispatcher side already reads these columns today.

Closes #373

### DIFF
--- a/packages/api/src/routes/v2/instances.ts
+++ b/packages/api/src/routes/v2/instances.ts
@@ -108,6 +108,28 @@ const createInstanceSchema = z.object({
     .nullable()
     .default(null)
     .describe('Debounce delay for group chats in milliseconds (null = use messageDebounceMinMs)'),
+  messageSplitDelayMode: z
+    .enum(['disabled', 'fixed', 'randomized'])
+    .default('randomized')
+    .describe('Between-chunk delay when agent responses are split into multiple messages'),
+  messageSplitDelayFixedMs: z
+    .number()
+    .int()
+    .min(0)
+    .default(0)
+    .describe('Fixed delay between split chunks in milliseconds (used when messageSplitDelayMode="fixed")'),
+  messageSplitDelayMinMs: z
+    .number()
+    .int()
+    .min(0)
+    .default(300)
+    .describe('Minimum delay between split chunks in milliseconds (used when messageSplitDelayMode="randomized")'),
+  messageSplitDelayMaxMs: z
+    .number()
+    .int()
+    .min(0)
+    .default(1000)
+    .describe('Maximum delay between split chunks in milliseconds (used when messageSplitDelayMode="randomized")'),
   agentGateEnabled: z.boolean().default(false).describe('Enable LLM response gate (pre-filter before agent dispatch)'),
   agentGateModel: z
     .string()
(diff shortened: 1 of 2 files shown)


vasconceloscezar:
## Review: SHIP

### Spec Compliance ✓

**Acceptance criteria verified:**

1. **Schema matches DB (4 fields, correct defaults)** ✓
   - `messageSplitDelayMode`: enum('disabled'|'fixed'|'randomized').default('randomized') ← matches DB ✓
   - `messageSplitDelayFixedMs`: number.int().min(0).default(0) ← matches DB ✓
   - `messageSplitDelayMinMs`: number.int().min(0).default(300) ← matches DB ✓
   - `messageSplitDelayMaxMs`: number.int().min(0).default(1000) ← matches DB ✓

2. **PATCH default-stripping** ✓
   - `updateInstanceSchema` explicitly overrides all 4 fields with `.optional()` (lines 222-225)
   - Prevents omitted PATCH fields from resetting DB values
   - Matches pattern used for `readReceipts`, `reactionAck`, etc.

3. **CLI flags consistent** ✓
   - Both create (310-313) and update (788-791) commands have flags
   - `--split-delay-mode`, `--split-delay-fixed`, `--split-delay-min`, `--split-delay-max`
   - Help text matches, parsing correct

4. **Helper wiring correct** ✓
   - \`applySplitDelayFields()\` mirrors \`applyDebounceFields()\` pattern
   - Called in \`buildInstanceBody()\` (line 155)
   - Uses \`setVal()\` for enum, conditional checks for numbers

5. **Validation passes** ✓
   - \`bun run typecheck\`: all 19 packages ✓
   - \`bunx biome check\`: 2 files, no issues ✓
   - \`bun test\`: 2982 pass, 0 fail ✓

### Code Quality ✓

- **Security**: Zod schemas enforce bounds and types
- **Correctness**: Field mapping correct, default-stripping prevents accidental resets
- **Scope**: Surgical—only split-delay fields, no refactoring
- **Maintainability**: Follows proven patterns, clear and consistent

### Summary

**All acceptance criteria met. Zero CRITICAL/HIGH findings.**

(Note: OpenAPI schema at \`packages/api/src/schemas/openapi/instances.ts\` doesn't document split-delay fields, but this appears intentional—it also doesn't document debounce fields. The actual API route returns all non-sensitive fields via \`sanitizeInstance()\`, so functionality is unaffected.)

**Status: ✅ READY TO SHIP**

PR #429, opened 2026-04-16
fix(cli): await stdout drain to prevent 64KB pipe truncation on --json

### LINKED ISSUE #402: CLI: `omni chats messages --json` truncates output at 64KB when piped
## Symptom

`omni chats messages <chat-id> --json` produces a complete JSON document when redirected directly to a file, but gets **truncated at exactly 65537 bytes** the moment its stdout is attached to any pipe (`| jq`, `| cat`, `| grep`).

## Reproduction

```bash
CHAT=<any chat with >64KB of message output, e.g., a busy group over --since 14d>

# Path A — direct redirect: full JSON, valid
omni chats messages "$CHAT" --since 14d --json 2>/dev/null > /tmp/a.json
wc -c /tmp/a.json          # 288806
jq 'length' /tmp/a.json    # 100

# Path B — through a pipe: truncated, invalid
omni chats messages "$CHAT" --since 14d --json 2>/dev/null | cat > /tmp/b.json
wc -c /tmp/b.json          # 65537
jq 'length' /tmp/b.json    # jq: parse error: Unfinished string at EOF

cmp /tmp/a.json /tmp/b.json
# /tmp/a.json /tmp/b.json differ: byte 65537, line 1514
```

The cutoff is **always 65537 bytes = 64KB + 1 byte**, which is the default Linux pipe buffer capacity. Stderr is empty.

## Likely root cause

Classic Node/Bun stdout flush bug: `process.stdout.write(bigString)` returns `false` (backpressure from the pipe's 64KB buffer). If the CLI exits before `'drain'` fires, everything past the first buffer is silently dropped. Direct file writes don't backpressure the same way, so they look fine.

Typical fix patterns:

```ts
// Option 1 — await drain
const ok = process.stdout.write(payload);
if (!ok) await new Promise((r) => process.stdout.once('drain', r));

// Option 2 — pre-serialize and write via a promise-friendly helper
await new Promise<void>((resolve, reject) =>
  process.stdout.write(payload, (err) => (err ? reject(err) : resolve())),
);
```

Both should be safe for small payloads and prevent truncation for large ones.

## Impact

Every pipe-based workflow (`omni chats messages ... --json | jq`, grep, tee, subsequent processing) silently loses data once the output crosses 64KB. No error is raised — the user only notices if they happen to pipe into a JSON-aware tool that complains about malformed input. For anything that just treats stdout as text, the data loss is invisible.

## Environment

- CLI: `@automagik/omni` (whatever's installed via `omni update`)
- Runtime: Bun
- OS: Linux 6.8.12-16-pve

## Acceptance

- Piped and redirected outputs are byte-identical for `omni chats messages --json` regardless of payload size.
- Consider auditing other `--json` commands for the same pattern (likely shared writer).

vasconceloscezar:
## Summary

- `omni chats messages <id> --json | cat > file` truncated at exactly 65537 bytes (64KB kernel pipe buffer).
- Switched JSON/raw output to `process.stdout.write(chunk, cb)` and await the callback through a `pendingStdoutWrites` set; `flushStdout()` now drains that set before exit.
- Added a regression test that spawns a child bun process, holds the reader idle for 100ms to trigger pipe backpressure, and asserts the full >100KB payload round-trips without truncation.

## Why this happened

`console.log` exposes no completion signal, and the previous `flushStdout()` called `process.stdout.write('', cb)` — in Bun the callback for an empty write can fire before the prior buffered data actually drains to the kernel, so the process exited with >64KB still sitting in Bun's internal stream buffer.

## Verification

- Before: `omni chats messages <large-id> --since 14d --json 2>/dev/null | cat > /tmp/b.json` → 65536 bytes; `| dd of=file` → 65536 bytes.
- After: same commands → 241040 bytes (complete), `jq length` parses to 100 messages.
- `make typecheck`, `make lint` (clean on changed files), `bun test` 263/263 pass in `@automagik/omni`, 3235/3235 repo-wide.

## Test plan

- [x] `bun test output.test` passes (11/11 including new regression test)
- [x] `bun test` repo-wide passes (3235/3235, pre-push hook)
- [x] Manual: `omni chats messages <large-id> --since 14d --json | wc -c` matches direct-redirect byte count
- [x] Manual: `omni chats messages <large-id> --since 14d --json | cat > /tmp/b.json && jq length /tmp/b.json` returns correct count

Closes #402

### DIFF
--- a/packages/cli/src/__tests__/output.test.ts
+++ b/packages/cli/src/__tests__/output.test.ts
@@ -97,8 +97,11 @@ describe('stdout pipe truncation (issue #402)', () => {
 
     try {
       const outputModulePath = join(import.meta.dir, '..', 'output.ts');
+      // Normalize backslashes on Windows so they don't become escape sequences
+      // when embedded in the single-quoted import literal below.
+      const importPath = outputModulePath.replace(/\\/g, '/').replace(/'/g, "\\'");
       const script = `
-import { list, flushStdout } from '${outputModulePath.replace(/'/g, "\\'")}';
+import { list, flushStdout } from '${importPath}';
 
 const items = Array.from({ length: 2000 }, (_, i) => ({
   id: 'msg-' + i,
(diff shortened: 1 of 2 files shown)


vasconceloscezar:
## Review: FIX-FIRST

**Acceptance Criteria: ✓ PASS**
All criteria verified for JSON 64KB truncation fix:
- [x] JSON paths use drain-aware writes via `process.stdout.write(callback)`
- [x] flushStdout() awaits all pending promises
- [x] Regression test simulates backpressure and validates >100KB round-trip

Tests: ✓ Pass (11/11) | Typecheck: ✓ Clean | Biome: ✓ Clean

---

## Code Quality: ✗ CRITICAL ISSUES FOUND

### [CRITICAL] Multiple unprotected exit paths bypass flushStdout()

**1. index.ts:605 — version check exits before Promise.all()**
```typescript
if (isRootVersionOnly) { console.log(...); process.exit(0); }
// then: await Promise.all([flushTelemetry(), flushStdout()])  ← never runs
```
Risk: `omni --json --version` can truncate

**2. output.ts:95 — error() function bypasses flush**
```typescript
export function error(message, details, exitCode): never {
  console.error(...);
  process.exit(exitCode);  // ← immediate exit, no flush
}
```
Risk: Any command with large JSON → output.error() = truncation

**3. Commands with direct process.exit() calls**
- update.ts: 8 calls (313, 351, 356, 361, 390, 398, 407, 419)
- providers-setup.ts:159, doctor.ts:759
All bypass index.ts:611 Promise.all flush

### [HIGH] Missing signal handlers
- No SIGINT/SIGTERM handlers; Ctrl+C exits without flush
- Common user pattern left unaddressed

### [HIGH] Non-JSON output scope unclear
- Table output still uses `console.log` (lines 201, 205, 211)
- Could hit same 64KB issue; PR scope not documented

### [MEDIUM] Test timing
- 100ms pause could flake on slow systems (mitigated by 30s timeout)

---

## Required Fixes

Before shipping:
1. **Critical:** Restructure version check to await flush before exit
2. **Critical:** Make error() async and await flush before process.exit()
3. **High:** Add SIGINT/SIGTERM signal handlers
4. **High:** Extend safety to table output or document JSON-only scope

The fix correctly solves issue #402 for happy-path commands, but critical exit vectors remain unprotected.

vasconceloscezar:
@gemini-code-assist — both gaps addressed in `844ef7b1`.

**Gap 1 (HIGH) — regression in non-JSON paths** — FIXED

You were right: the prior patch only tracked JSON/raw writes via `pendingStdoutWrites`, so human-format paths (tables, success/warn/info, key/value, plain `data()`) still hit `console.log` and lost the best-effort empty-string drain they relied on before. Large human tables piped through a slow reader would have truncated at exactly 64KB.

Fix in `packages/cli/src/output.ts`:
- Every stdout emission now routes through `writeStdoutLine`, so every format (JSON, raw, table, success, warn, info, header, dim, keyValue, printObject, printTable header/separator/rows, scalar `data`) gets drain tracking.
- Stderr paths (`error`, JSON-mode `warn`/`info`) stay on `console.error`.
- Added belt-and-suspenders in `flushStdout()`: after awaiting tracked writes, a trailing `await new Promise((r) => process.stdout.write('', r))` gives any untracked stdout bytes a final drain opportunity.

**Gap 2 (MEDIUM) — Windows portability** — FIXED

Applied your suggested normalisation in `packages/cli/src/__tests__/output.test.ts`:
```ts
const importPath = outputModulePath.replace(/\\/g, '/').replace(/'/g, "\\'");
```
Also added a second regression test that forces `OMNI_FORMAT=human` + `NO_COLOR=1` through the same 100ms slow-pipe scenario and asserts the full 100KB+ human-formatted table lands intact (including the header row and `msg-1999`).

**Validation** — 264 CLI tests pass; 3236 tests pass overall; `tsc --noEmit -p packages/cli/tsconfig.json` clean; `biome check` on the two edited files clean.

Commit: `844ef7b19d8d4b0ef37aa52f625b149f1dd5fd34`

PR #453, opened 2026-04-19
fix(api): auto-set reply filter when agent assigned to instance

### LINKED ISSUE #443: fix: auto-set reply filter to 'all' + onDm when agent is assigned to instance
## Problem

When assigning an agent to an instance via `omni instances update --agent-fk-id <uuid>`, the `agentReplyFilter` remains `null`. This silently causes the instance to **receive messages but never dispatch them to the agent** — no error, no warning, just silent drops.

**CLI version:** `2.260410.1`

### Reproduction

```bash
# Create instance + agent + provider (all good)
omni instances create --name my-bot --channel telegram --telegram-token <token>
omni providers create --name my-provider --schema nats-genie --base-url nats://localhost:4222 ...
omni agents create --name my-agent --provider custom --agent-provider <uuid>

# Wire agent to instance
omni instances update <instance-id> --agent-fk-id <agent-uuid>

# Send message to bot → message.received event appears, but NO dispatch to agent
# Because agentReplyFilter is null
```

### Root cause

`agentReplyFilter: null` means "no reply filter configured" which the dispatch pipeline interprets as "don't reply to anything." The user expects that assigning an agent means "this agent should respond to messages."

### Expected behavior

When an agent is assigned to an instance (via `--agent-fk-id`, `--agent`, or during `instances create`), the API should **auto-set a sensible default reply filter** if none is explicitly configured:

```json
{
  "mode": "all",
  "conditions": {
    "onDm": true
  }
}
```

This matches what every user expects: "I assigned an agent, it should respond."

### Workaround

Manually set after agent assignment:
```bash
omni instances update <id> --reply-filter-mode all --reply-on-dm
```

But this is a hidden step that causes silent failures — the worst kind of bug.

### Acceptance criteria

- [ ] When `agentId` is set (via create or update) and `agentReplyFilter` is null, auto-populate with `{ mode: "all", conditions: { onDm: true } }`
- [ ] If user explicitly sets `--reply-filter-mode` during the same command, respect their choice (don't override)
- [ ] Log a warning if agent is assigned with no reply filter: `"Agent assigned but no reply filter set — defaulting to mode:all"`
- [ ] Add to `omni instances create --help` docs: note that reply filter defaults to "all" when agent is assigned

### Impact

This caused a complete silent failure during a Telegram bot setup — 3 messages sent, 0 dispatched, 0 errors logged. Took ~30 minutes of debugging to identify that `agentReplyFilter: null` was the blocker.

vasconceloscezar:
## Summary

- POST /instances: when `agentId` is set and `agentReplyFilter` is `undefined`, auto-assign the default filter `{ mode: 'all', conditions: { onDm: true, onMention: true, onReply: true, onNameMatch: false } }`.
- PATCH /instances/:id: reuse the existing `current` fetch, compute `finalAgentId` from payload or persisted record, and apply the same default only when neither payload nor persisted record already has a filter.
- Explicit `null` from the caller is preserved (opt-out stays respected).

Closes #443

## Why

Assigning an agent without a reply filter silently dropped all inbound messages — a `null` filter was interpreted downstream as "reply to nothing." This caused ~30 minutes of invisible failures during Telegram setup.

## Test plan

- [x] `bun run typecheck` in `packages/api` — passes
- [x] `bun test` in `packages/api` — 625 pass / 0 fail
- [x] Pre-push full workspace suite — 3150 pass / 0 fail
- [ ] Manual: create instance with `--agent-fk-id <uuid>` → verify `agentReplyFilter` is auto-populated with default object and messages dispatch
- [ ] Manual: patch instance to add agent (no filter payload) → verify default filter is stored
- [ ] Manual: patch instance with `agentReplyFilter: null` → verify explicit null is preserved

### DIFF
--- a/packages/api/src/routes/v2/instances.ts
+++ b/packages/api/src/routes/v2/instances.ts
@@ -531,6 +531,14 @@ instancesRoutes.post('/', zValidator('json', createInstanceSchema), async (c) =>
   const services = c.get('services');
   const channelRegistry = c.get('channelRegistry');
 
+  if (data.agentId && data.agentReplyFilter === undefined) {
+    data.agentReplyFilter = {
+      mode: 'all',
+      conditions: { onDm: true, onMention: true, onReply: true, onNameMatch: false },
+    };
+    log.info('Auto-set agentReplyFilter for new instance', { agentId: data.agentId });
+  }
+
   // Map generic `token` → channel-specific DB column + resolve connect token
   const connectToken = resolveChannelToken(data);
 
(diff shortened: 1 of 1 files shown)


namastex888:
Superseded by PR #484 (merged 2026-04-22), which ships the same auto-set reply filter behavior with 12 unit tests. Closing this older duplicate PR. Thanks for the original contribution — much of the design intent is reflected in the merged fix.

Issue #536, opened 2026-04-24
chatId normalization inconsistent between agent-dispatcher and follow-up sweeper (WhatsApp @lid / @s.whatsapp.net)


vasconceloscezar:
## Context

When a WhatsApp chat arrives through Baileys, Omni carries two forms of the chat identifier:

| Field | Example |
|---|---|
| `chats.external_id` | `35450727202836@lid` |
| `chats.canonical_id` | `5519988404000@s.whatsapp.net` |

The **agent-dispatcher** path (real inbound message → seller) strips the JID suffix before handing off to `agentRunner.run({ chatId, ... })`. So downstream agno agents compute `sessionId = chatId` = `35450727202836` (stripped, no suffix).

The **follow-up sweeper** path emits `chat.idle_timeout` events with `payload.chatId = chats.external_id` = `35450727202836@lid` (raw, WITH suffix).

Both paths reach the same agno agent via `call_agent` → `agentRunner.runOrStream` → `computeSessionId('per_chat', ...)` (which returns `chatId`). Because the chatIds disagree, the two paths target different session rows in agno.

## Impact

- The seller builds history under `35450727202836` (stripped).
- Every follow-up `chat.idle_timeout` event looks up `35450727202836@lid` (raw).
- Downstream agno side finds no session → the follow-up flow can't reconstruct context from the seller's real history.

Observed in production 2026-04-24 19:39:30 UTC and reproducible 3/3 times across a real WhatsApp chat — not intermittent.

This was also flagged as "Bug 2" in the PR body of namastexlabs/genie-hv-eugenia#60 ("session-id mismatch between Omni and agno — some real follow-ups hit `session is None or has no session_data` at precheck (1 of 3 observed)") and has now been fully isolated.

## Expected

`agent-dispatcher` and `follow-up sweeper` should emit the same `chatId` (and therefore same `sessionId`) for a given chat on a given instance. Pick one:

- (a) Both paths strip JID suffixes (preferred — matches what the dispatcher already does today and preserves existing session history).
- (b) Both paths preserve the raw `external_id` (more invasive — existing agno session rows would be orphaned and need migration).

## Where to look

- `packages/api/src/plugins/agent-dispatcher.ts` — the seller's chatId resolution path (line ~953 / ~1948 is where `agentRunner.run({ chatId })` gets called).
[...]

Issue #528, opened 2026-04-24
follow-up automation fires after handoff — race between chat.handoff_activated and chat.idle_timeout consumer


namastex888:
## Summary

Follow-up messages from automations triggered by `chat.idle_timeout` keep being delivered to leads **after** a handoff has been issued via `POST /api/v2/messages/send/handoff`. Expected behavior: once a chat is handed off (settings.agentPaused=true), the follow-up pipeline must go silent until an operator resumes the agent.

Confirmed in production (Hapvida/Eugenia-seller on Gupshup instance). 6 out of 184 handoffs in a 48h window (≈3.3%) received at least one follow-up after the handoff fired. In one case a lead received **6 follow-ups** over the following 12h after being handed off.

## Impact

- Lead receives contradictory messages: a handoff ("vou te conectar com um consultor") immediately followed by automated follow-ups from the same bot persona.
- Breaks operator trust — the human consultant picks up the conversation and finds the bot has kept poking the lead.
- Not rare: hits any lead that gets handed off while the follow-up sweeper had a due/queued event.

## Architecture (for context)

Two paths interact:

1. **Handoff path** — agent calls `human_handoff` tool → `POST /api/v2/messages/send/handoff`:
   - `plugin.sendMessage(HANDOFF)` sends the goodbye message.
   - `services.chats.update({ settings: { agentPaused: true } })` → emits `chat.handoff_activated` (see `packages/api/src/services/chats.ts:603-629`).
   - `follow-up-hooks.ts` NATS consumer receives `chat.handoff_activated` → calls `followUpLifecycle.disarm({ reason: 'handoff' })` → row in `chat_follow_up_state` gets `disarmReason='handoff'`.

2. **Follow-up path** — sweeper ticks every 15s (`services/follow-up-sweeper.ts::findAndLockDue`):
   - `SELECT ... FROM chat_follow_up_state WHERE disarmReason IS NULL AND nextFireAt <= now() FOR UPDATE SKIP LOCKED`.
   - For each due row, emits `chat.idle_timeout` into NATS SYSTEM stream.
   - Automation engine consumer (`automation-engine-chat-idle_timeout`, durable, `maxRetries=3`) consumes the event and runs the matching automation (e.g. `Eugenia follow-up (Haiku)` = call_agent + send_message).

Both paths are correct in isolation. The bug is in the **gap between them**.

## Root cause

There are at least two concurrent race windows:

### Race #1 — event already in the NATS queue

[...]

Issue #515, opened 2026-04-23
agno 2.5 id migration — follow-up: SDK client + CLI providers commands still read legacy agent_id/team_id/workflow_id


vasconceloscezar:
## Context

Follow-up to #509 / PR #511.

PR #511 fixed `packages/core/src/providers/agno-client.ts` and its shared types so `AgnoClient.discover()` now populates `id` correctly under agno 2.5+. It explicitly left two other surfaces out of scope because they use **separate type copies** (not reusing `core/providers/types.ts`), so the fix does not propagate.

This issue tracks those remaining surfaces so they don't get lost.

### 1. SDK client type copies

`packages/sdk/src/client.ts:421,432,443` — the SDK's `AgnoAgent` / `AgnoTeam` / `AgnoWorkflow` declarations still use the pre-2.5 names (`agent_id` / `team_id` / `workflow_id`). Any SDK consumer listing agno entities today sees `undefined` in the ID field.

### 2. CLI providers commands

`packages/cli/src/commands/providers.ts:400,425,451` — `omni providers agents list` / `teams list` / `workflows list` read the legacy fields when rendering the table. On agno 2.5 the ID column is blank.

## Fix pattern

Mirror PR #511:

1. Update type declarations (in each package's own `types` or inline definitions) to declare `id: string` as the authoritative field, keep legacy `agent_id` / `team_id` / `workflow_id` as optional + deprecated for backward compat with pre-2.5 deployments.
2. Where values are read, use `e.id ?? e.agent_id` (analogously for team/workflow) so both shapes work.
3. Add one regression test per package exercising the agno 2.5 shape and asserting `id` is defined.

## Acceptance criteria

- [ ] `packages/sdk/src/client.ts` — agent/team/workflow ID readers handle both `id` and legacy `{agent|team|workflow}_id`.
- [ ] `packages/cli/src/commands/providers.ts` — `omni providers agents list` / `teams list` / `workflows list` show the correct ID on agno 2.5 and remain compatible with pre-2.5 deployments.
- [ ] Regression tests added in SDK + CLI packages mirroring the `agno-client.test.ts` style.
- [ ] Optional: factor the three `AgnoAgent` / `AgnoTeam` / `AgnoWorkflow` type declarations into one shared module so this class of drift doesn't recur.

## References

- Parent issue: #509
- Merged fix (core): #511
- Surfaces flagged in #511 body, "Follow-up (out of scope for this PR)" section.

vasconceloscezar:
[...]

Issue #509, opened 2026-04-23
agno-client: discover() reads missing field `agent_id` — agno 2.5 schema uses `id`


vasconceloscezar:
## Context

Discovered while wiring a second agno agent (`eugenia-followup`) on an existing instance. Manual registration via SQL worked, but `omni` discovery would not have picked up either agent correctly.

## Repro

```bash
curl -H "Authorization: Bearer $TOKEN" http://<agno-baseurl>/agents | jq '.[0] | keys'
```

Returns:
```json
["db_id", "id", "is_component", "memory", "model", "name", "response_settings", "sessions", "system_message", "tools"]
```

No `agent_id` top-level key. Agno schema (`agno/os/routers/agents/schema.py:19`) is:
```python
class AgentResponse(BaseModel):
    id: Optional[str] = None
    name: Optional[str] = None
    ...
```

## Bug

`packages/core/src/providers/agno-client.ts:102`:

```ts
...agents.map((a) => ({
  id: a.agent_id,   // ← undefined for every agno 2.5+ agent
  name: a.name,
  type: 'agent' as const,
  description: a.description,
  metadata: a.model ? { model: a.model } : undefined,
})),
```

`a.agent_id` does not exist on the agno response. `id` is always `undefined` for every discovered agno agent.

## Impact

- Any Omni UI / flow that lists or picks agno agents via the provider's `discover()` sees empty IDs.
- Auto-registration of agno agents (populating `agents.metadata.providerAgentId` from discovery) cannot work.
- Runtime dispatch is unaffected for anyone who registered manually (we used direct SQL: `UPDATE agents SET metadata = '{"providerAgentId":"eugenia-followup"}' WHERE id = ...`).

## Fix

`agno-client.ts:102`:

```diff
- id: a.agent_id,
+ id: a.id ?? a.agent_id,
```

or simply:

```diff
- id: a.agent_id,
+ id: a.id,
```

if agno 2.x is the only supported version. The type declaration of `AgnoAgent` upstream should also be updated: the field is `id: string`, not `agent_id`.

Similar check worth doing for `AgnoTeam` and `AgnoWorkflow` in the same file (`a.team_id`, `a.workflow_id`) — agno 2.5 may have unified those to `id` too.

## Notes

- agno version confirmed: `agno[os]>=2.5,<3.0`.
- No runtime regression for existing integrations that bypass discovery.
- Discovered while resolving #502 (provider-id flag UX issue) — the two bugs are independent.

[...]

PR #538, opened 2026-04-24
fix(handoff): relax Gupshup-only gate — pause + disarm on any channel (#537)

### LINKED ISSUE #537: handoff: relax Gupshup-only gate — agent pause + follow-up disarm should work on any channel
## Problem

`POST /messages/send/handoff` returns **HTTP 400 `Handoff is only supported on Gupshup instances`** for every non-Gupshup channel. As a result, agents running on `whatsapp-baileys`, `discord`, `telegram`, `slack` etc. have no way to activate handoff — which in turn means they cannot pause the agent or disarm the follow-up sweeper for a chat, even when the lead explicitly asks for a human.

Observed in production on instance `testonho` (channel `whatsapp-baileys`):

- Lead typed "Quero falar com um humano"
- Eugenia seller called its `human_handoff` tool → hit `POST /messages/send/handoff`
- Omni rejected with 400 `Handoff is only supported on Gupshup instances`
- Seller replied *"Um consultor vai entrar em contato com você em breve"* but **nothing happened on Omni's side**
- Follow-up sweeper kept firing → 2 more follow-ups reached the user in the next 5 minutes asking "você conseguiu falar com nosso pessoal?"

Agno session (for context): `217046273028329`, run `8ddfc405-568f-43ff-b9a1-395d622f619b`, tool result literally `{"ok": false, "error": "HTTP 400: {\"error\":\"Handoff is only supported on Gupshup instances\"}"}`.

## The blocker is narrower than it looks

The surrounding infrastructure is already channel-agnostic. It's only the HTTP entry that enforces Gupshup:

- `packages/api/src/routes/v2/messages.ts:1469` — the hardcoded check:
  ```ts
  if (instance.channel !== 'gupshup') {
    return c.json({ error: 'Handoff is only supported on Gupshup instances' }, 400);
  }
  ```
- Everything downstream is channel-neutral:
  - `packages/api/src/services/chats.ts:587-632` — `chats.update({settings: {agentPaused: true}})` detects the false → true transition and emits `chat.handoff_activated` regardless of channel.
  - `packages/api/src/plugins/follow-up-hooks.ts:126-137` — subscribes to `chat.handoff_activated`, calls `followUpLifecycle.disarm({reason: 'handoff'})`.
  - `packages/api/src/plugins/agent-dispatcher.ts:2565-2575` — checks `settings.agentPaused === true` and skips agent dispatch (channel-agnostic).
  - `packages/api/src/services/follow-up-sweeper.ts:125` — SQL excludes `agentPaused = true` from sweeper ticks.

What Gupshup adds is a channel-specific side effect: it sends a `msg_type: HANDOFF` payload to Gupshup so the ticket pops up in their agent UI (`messages.ts:1481-1505`). Other channels have no equivalent protocol. But the agent-pause + disarm side of the chain does not need that payload — those are controlled entirely by `chats.settings.agentPaused`.

## Proposal

Split the HTTP route into two behaviors, keyed on whether the channel plugin advertises handoff support:

1. **Channel has native handoff** (currently only Gupshup) → keep today's path: send channel-specific HANDOFF payload via plugin, flip `agentPaused`, emit event, audit row in `handoffLogs`.
2. **Channel has no native handoff** → skip the plugin send, still flip `agentPaused: true`, still insert an audit row (with `channel_handoff_supported: false`), and rely on the agent to send a user-facing farewell via normal `send_message`.

### Suggested API shape (minimal drift)

Keep `POST /messages/send/handoff` as the single entry. Add a channel capability probe:

```ts
// packages/core/src/plugins/channel.ts (new method on the plugin interface, optional)
interface ChannelPlugin {
  /** Whether this channel supports a native handoff protocol (msg_type: HANDOFF etc.). */
  supportsHandoff?(): boolean;
}
```

Then:

```ts
const plugin = channelRegistry.get(instance.channel as ChannelType);
const hasNativeHandoff = plugin?.supportsHandoff?.() === true;

if (hasNativeHandoff) {
  // Existing Gupshup path — plugin.sendMessage with isHandoff metadata
  const result = await plugin.sendMessage(...);
  handleSendResult(...);
}

// Channel-agnostic side effects — run for every channel
await services.chats.update(data.chatId, {
  settings: { agentPaused: true },
});
await services.followUpLifecycle.disarm({ chatId, instanceId, reason: 'handoff' });
await db.insert(handoffLogs).values({
  ...,
  channelHandoffSupported: hasNativeHandoff,
});
```

Drop the hardcoded `instance.channel !== 'gupshup'` check entirely.

### Alternative (narrower) — introduce a second endpoint

If mixing the two behaviors on one route feels wrong, add `POST /messages/send/handoff/soft` (or `.../pause`) that only flips `agentPaused` + disarms follow-up + writes the audit row, without any channel-specific send. The seller's `human_handoff` tool then picks between the two based on the channel capability it can detect.

I mildly prefer the single-endpoint shape — it keeps the contract the seller calls stable across channels, and the capability probe lives inside Omni where it belongs.

## Acceptance criteria

- [ ] `POST /messages/send/handoff` succeeds (HTTP 200 or 201) for at least `whatsapp-baileys` and `discord`; for each it:
  - [ ] Sets `chats.settings.agentPaused = true` (false → true transition).
  - [ ] Emits `chat.handoff_activated` exactly once.
  - [ ] Disarms the active follow-up sequence for the chat (row disappears from `chat_follow_up_state` sweeper output).
  - [ ] Blocks subsequent agent dispatch for that chat (`agent-dispatcher.ts:2571` gate).
  - [ ] Writes a row into `handoff_logs` with `channel_handoff_supported` truthfully populated.
- [ ] Gupshup path is unchanged: native HANDOFF payload still goes out via plugin.
- [ ] For channels lacking native handoff, no channel-level send happens; the route does NOT 400 and the user gets no spurious "system" message from Omni (the agent handles the farewell message).
- [ ] Integration tests cover both paths on at least one channel each.

## Repro (for whatever channel you pick to fix)

1. Register a whatsapp-baileys (or discord) instance with an agent.
2. Have the lead type a handoff trigger.
3. Agent invokes its handoff tool → hits `POST /messages/send/handoff`.
4. Expect: 200; `chat.agentPaused=true`; zero follow-ups delivered afterward; agent no longer responds to new messages until resumed.

## Related

- Observed in `namastexlabs/genie-hv-eugenia` (Eugenia seller on `testonho` / whatsapp-baileys). Agno session with full tool-error trace available on request.
- Similar to past upstream issues opened in this series: #502, #509, #515, #536 — all surfaced while wiring the Eugenia follow-up agent.

vasconceloscezar:
Closes #537

## Problem

`POST /messages/send/handoff` was hardcoded to return **HTTP 400** for every non-Gupshup channel, blocking agents on `whatsapp-baileys`, `discord`, `telegram`, and `slack` from pausing themselves or disarming the follow-up sweeper when a lead asked for a human. Observed in production: Eugenia seller on a `whatsapp-baileys` instance (`testonho`) called its `human_handoff` tool, got 400 back, replied a farewell anyway — but **nothing paused on Omni's side**, so the follow-up sweeper delivered two more nag messages in the next 5 minutes.

## Root cause

The surrounding infrastructure is already channel-agnostic (`chats.update` emits `chat.handoff_activated` regardless of channel; `follow-up-hooks` disarms; `agent-dispatcher` skips dispatch when `agentPaused=true`; sweeper excludes paused rows via #528). Only the HTTP entry enforced Gupshup:

```ts
// packages/api/src/routes/v2/messages.ts:1469
if (instance.channel !== 'gupshup') {
  return c.json({ error: 'Handoff is only supported on Gupshup instances' }, 400);
}
```

## Fix — capability-based branching, single endpoint

1. Added `canHandoff?: boolean` to `ChannelCapabilities` (`packages/channel-sdk/src/types/capabilities.ts`). Channels that advertise a native HANDOFF protocol set it `true`; everyone else leaves it `undefined`/`false`.
2. Set `canHandoff: true` in `GUPSHUP_CAPABILITIES` — only channel today with a native `msg_type: HANDOFF` protocol.
3. Route rewrite in `packages/api/src/routes/v2/messages.ts`:
   - Removed the hardcoded `instance.channel !== 'gupshup'` 400.
   - Uses `plugin.capabilities?.canHandoff === true` to decide whether to run `plugin.sendMessage` with `isHandoff` metadata.
   - **Always** runs the channel-agnostic side effects: `chats.update({agentPaused: true})`, `followUpLifecycle.disarm({reason: 'handoff'})`, `handoffLogs` audit insert.
   - Stores `channelHandoffSupported` in the audit row's `metadata` JSONB so downstream tooling can tell native from soft handoffs.
   - Response: `status: 'sent'` when the channel pushed a native payload, `status: 'paused'` when only the agent-pause chain ran.

No schema change — audit metadata lives in the existing JSONB column.

## Diff

```
packages/channel-sdk/src/types/capabilities.ts   | +11
packages/channel-gupshup/src/capabilities.ts      | +1
packages/api/src/routes/v2/messages.ts            | +28 -22
```

## Validation

- Typecheck: `@omni/api` + `@omni/channel-sdk` + `@omni/channel-gupshup` all green.
- Biome: clean on all 3 changed files.
- Tests on affected packages: **1169 pass / 0 fail / 271 skip**.
- Full pre-push suite will run on push (repo gate).

## Acceptance criteria (from #537)

- [x] `POST /messages/send/handoff` no longer 400s for `whatsapp-baileys` / `discord` / `telegram` / `slack`.
- [x] For channels **without** native handoff: sets `agentPaused=true`, emits `chat.handoff_activated`, disarms follow-up, writes `handoff_logs` row with `channelHandoffSupported: false`, and skips any channel-level send (agent handles the farewell).
- [x] For channels **with** native handoff (Gupshup): unchanged — native HANDOFF payload still goes out via plugin; audit row now records `channelHandoffSupported: true`.
- [x] Blocks subsequent agent dispatch via existing `settings.agentPaused` gate in `agent-dispatcher.ts`.
- [ ] Integration tests on both paths — **deferred**. No existing route harness for `/send/handoff` (no `messages.test.ts`); adding one is out of scope for this fix. Logged as follow-up if worth adding. All downstream machinery (`chats.update` hook, follow-up-hooks consumer, sweeper filter, dispatcher gate) is already covered by existing unit/integration tests.

## Related

- #502 — CLI provider-id flag UX (closed)
- #509 / #511 — agno-client `agent_id` → `id` migration (merged)
- #515 / #516 — agno 2.5 migration propagated to SDK + CLI (merged)
- #528 / #535 — follow-up disarm race on handoff (merged — the sweeper filter and synchronous disarm added there are what makes this fix safe on non-native channels)
- #536 — referenced as prior art in the series

### DIFF
--- a/packages/api/src/routes/v2/messages.ts
+++ b/packages/api/src/routes/v2/messages.ts
@@ -1484,6 +1484,12 @@ messagesRoutes.post('/send/handoff', zValidator('json', sendHandoffSchema), asyn
     });
   }
 
+  // Resolve the recipient for every channel — even when we won't push a
+  // native payload we still want the audit row to record a real platform
+  // identifier (phone/JID) rather than the caller's input, which may be
+  // an Omni Person UUID. See issue #537 + gemini review on #538.
+  const resolvedTo = await resolveRecipient(data.to, instance.channel, services);
+
   // Channels that declare `canHandoff: true` receive a channel-specific
   // HANDOFF payload (currently only Gupshup). For every other channel the
   // route still runs the channel-agnostic side effects below — agentPaused,
@@ -1494,7 +1500,6 @@ messagesRoutes.post('/send/handoff', zValidator('json', sendHandoffSchema), asyn
 
   let channelSendResult: Awaited<ReturnType<typeof plugin.sendMessage>> | null = null;
   if (hasNativeHandoff) {
-    const resolvedTo = await resolveRecipient(data.to, instance.channel, services);
     const outgoingMessage: OutgoingMessage = {
       to: resolvedTo,
       content: { type: 'text', text: data.text } as OutgoingContent,
@@ -1533,8 +1538,8 @@ messagesRoutes.post('/send/handoff', zValidator('json', sendHandoffSchema), asyn
     .values({
       instanceId: data.instanceId,
       chatUuid: data.chatId, // chatId in this route is the DB UUID of the chat
-      chatId: data.to, // raw phone/JID used as chat identifier on the channel
-      toPhone: data.to,
+      chatId: resolvedTo, // resolved platform identifier (phone/JID)
+      toPhone: resolvedTo,
       text: data.text,
       extraInfo: data.dadosLead ?? data.extraInfo ?? null,
       agentId: instance.agentId ?? null,
--- a/packages/api/src/routes/v2/messages.ts
+++ b/packages/api/src/routes/v2/messages.ts
@@ -1466,10 +1466,6 @@ messagesRoutes.post('/send/handoff', zValidator('json', sendHandoffSchema), asyn
 
   const instance = await services.instances.getById(data.instanceId);
 
-  if (instance.channel !== 'gupshup') {
-    return c.json({ error: 'Handoff is only supported on Gupshup instances' }, 400);
-  }
-
   if (!channelRegistry) {
     throw new OmniError({
       code: ERROR_CODES.CHANNEL_NOT_CONNECTED,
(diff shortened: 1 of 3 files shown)


vasconceloscezar:
Addressed the gemini review finding in d32502cf. resolveRecipient moved above the canHandoff branch; handoff_logs.chatId and toPhone now use resolvedTo for every channel (pre-existing inconsistency on the Gupshup path fixed incidentally).

PR #539, opened 2026-04-24
fix(follow-up): harden sweeper agentPaused SQL + test cleanup (gemini review on #535)

vasconceloscezar:
Follow-up to PR #535 / issue #528 — addresses the gemini-code-assist review left after merge.

## HIGH — SQL cast risk

The sweeper's `findAndLockDue` predicate was:

```sql
(chats.settings->>'agentPaused')::boolean IS DISTINCT FROM true
```

The explicit boolean cast raises a runtime error if `settings.agentPaused` ever holds a non-boolean text value (empty string, `\"1\"`, stray type drift, etc.), which would kill the entire sweeper tick and silently break follow-ups platform-wide. In practice we always write booleans today, but the risk is latent — a bad migration or a stray external write would take everything down.

Replaced with a null-safe text comparison:

```sql
(chats.settings->>'agentPaused') IS DISTINCT FROM 'true'
```

Postgres's `->>` returns NULL for missing keys and `'true'`/`'false'` for booleans, so this is null-safe, matches both boolean `true` and JSON string `\"true\"`, and never raises at runtime.

## MEDIUM — Test settings cleanup

The agentPaused regression test's cleanup overwrote the whole `settings` JSONB with `{}`, which would clobber any unrelated keys another test might rely on. Swapped to use the JSONB `-` operator so only the key we added is removed:

```ts
await db
  .update(chats)
  .set({ settings: sql\`\${chats.settings} - 'agentPaused'\` })
  .where(eq(chats.id, testChatId));
```

## Validation

- Typecheck: `@omni/api` green.
- Biome: clean on both changed files.
- Bun test locally: 0 pass / 8 skip / 0 fail (integration tests gate on `ENABLE_DB_TESTS`, same as before).

## Diff

```
packages/api/src/services/follow-up-sweeper.ts    | +10 -2
packages/api/src/__tests__/follow-up-sweeper.test.ts | +5 -2
```

## Related

- PR #535 / issue #528 — original fix
- gemini-code-assist[bot] review on #535

### DIFF
--- a/packages/api/src/__tests__/follow-up-sweeper.test.ts
+++ b/packages/api/src/__tests__/follow-up-sweeper.test.ts
@@ -83,6 +83,13 @@ describeWithDb('FollowUpSweeperService (integration)', () => {
 
   afterEach(async () => {
     await db.delete(chatFollowUpState).where(eq(chatFollowUpState.chatId, testChatId));
+    // Defensive cleanup — if a test wrote `settings.agentPaused`, strip only
+    // that key. Runs unconditionally so a failed test can't leak paused
+    // state into subsequent cases sharing `testChatId`.
+    await db
+      .update(chats)
+      .set({ settings: sql`${chats.settings} - 'agentPaused'` })
+      .where(eq(chats.id, testChatId));
   });
 
   afterAll(async () => {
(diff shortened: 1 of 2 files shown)


vasconceloscezar:
Addressed the gemini review in be5fe95a. Setup now merges via `jsonb_set(COALESCE(settings, '{}'::jsonb), ...)` instead of overwriting, and the cleanup moved into `afterEach` so it runs unconditionally.

PR #544, opened 2026-04-27
feat(sdk): rename @omni/sdk → @automagik/omni-sdk + drop dispatcher-retry mock + instructive pre-push

namastex888:
## Summary

Two related shifts that landed together because the pre-push gate
surfaced the second one when I tried to push the first:

1. **Rename `@omni/sdk` → `@automagik/omni-sdk`** so the SDK can be
   published to the public npm registry under the @automagik scope.
   The SDK at packages/sdk has always been the public-facing client
   that other apps (notably @khal-os/brain) consume — but its
   workspace-local @omni/* namespace blocked publish, so external
   installs of brain hit a 404 and fell back to a "@omni/sdk
   unavailable" wizard error.

2. **Drop `mock.module('@omni/core', ...)` from
   agent-dispatcher-retry.test.ts** + spread the real surface in
   agent-dispatcher.test.ts + rewrite the pre-push hook to be
   instructive. Pre-push surfaced 13 Logger test failures and 3
   typecheck errors when I tried to push (1); per repo policy ("we
   always fix things WHEN we find") I fixed both rather than bypass.

## Changes — Rename

- `packages/sdk/package.json`: name → `@automagik/omni-sdk`,
  `publishConfig.access = public`.
- 62 src files across `packages/cli` + `apps/ui` updated to import
  `@automagik/omni-sdk`. Workspace `:*` pin preserves internal
  resolution, no version drift.
- READMEs updated for consistency.
- biome import-order auto-fix on `packages/cli/src/commands/providers.ts`.

**No breaking changes outside this monorepo.** The only external
consumer of `@omni/sdk` was `@khal-os/brain` via a local `file:` link,
which will swap to a real npm dep once this lands and `npm publish`
lifts `@automagik/omni-sdk@2.260423.2` to the registry.

## Changes — Test infra & pre-push hook

- **`agent-dispatcher-retry.test.ts`**: removed the
  `mock.module('@omni/core', ...)` call. Bun's mock.module is
  process-wide; the truncated factory return value was poisoning
  Logger tests in the full suite. Even with a full-surface spread
  (`require('@omni/core')` + ...real) the contamination persisted —
  the dispatcher tests do not assert on logger output, so the
  surgical fix is to drop the mock entirely. 28 dispatcher-retry
  tests still pass.
- **`agent-dispatcher.test.ts`**: the @omni/core mock factory now
  spreads `require('@omni/core')` over overrides so real classes
  (LogBuffer, configureLogging, OmniError) keep their shape. The
  comment that claimed Bun's mock.module had "merge behavior" was
  wrong; corrected inline.
- **`.husky/pre-push`**: rewritten to be instructive. Previous version
  exited silently with `set -e`, leaving agents guessing. New version
  has a 38-line header explaining the contract, step labels with
  remediation hints printed inline ("If errors reference SDK types,
  run `cd packages/sdk && bun run build`"), and a final ✓
  acknowledgment when all gates pass.

## Validation

- `make typecheck` ✅ clean (was 3 pre-existing errors in providers.ts,
  fixed by SDK rebuild during the rename).
- `make lint` ✅ clean.
- `bun test` ✅ **3801 pass / 271 skip / 0 fail** (was 13 fail on dev
  before this branch).
- `bun install` ✅ workspaces relinked cleanly.

## Follow-up

Brain repo PR will swap its file:-link dependency for the published
@automagik/omni-sdk once a tagged release lifts it to the registry.

## Context

Code-freeze + dog-food season. Felipe's directive when pre-push
flagged pre-existing failures: "we are not in favor of 'this was
already here', we always fix things WHEN we find. update the hook
output to be more instructive." Both done.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,5 +1,45 @@
 set -e
 
+# ============================================================================
+# pre-push gate: typecheck + tests must be GREEN before any push leaves here.
+#
+# Why so strict? Code freeze + dog-food season. Every coding agent that pushes
+# from this repo inherits the contract: zero typecheck errors, zero failing
+# tests. If you (human or agent) hit a failure here, FIX IT — even if the
+# failure was already on dev. "It was already broken" is not an exception;
+# the hook stays strict so the repo state monotonically improves.
+#
+# Common causes + how to fix them:
+#
+#   • Typecheck error in packages/cli referencing SDK types
+#       → SDK dist/ is stale. Run `cd packages/sdk && bun run build`,
+#         then re-push. The hook auto-builds SDK if dist/ is missing
+#         entirely, but a stale dist (older than source) is harder to
+#         detect — always rebuild after touching packages/sdk/src.
+#
+#   • Logger / Buffer / formatter tests failing in full suite but
+#     passing in isolation
+#       → Some other test file is calling `mock.module('@omni/core', ...)`
+#         with a truncated surface. Bun's mock.module is process-wide,
+#         not file-scoped — the truncated mock contaminates every other
+#         test that resolves @omni/core. Drop the mock.module call from
+#         the dispatcher / api test (the tests usually do not assert on
+#         logger output anyway), or spread `require('@omni/core')` over
+#         the overrides so unrelated modules keep their real exports.
+#
+#   • Unrelated test failures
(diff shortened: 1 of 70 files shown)


namastex888:
Refactoring per Felipe — subpath export in @automagik/omni instead of separate package. New PR incoming.

Issue #445, opened 2026-04-18
bug(events): automation engine subscriptions silently die — ephemeral NATS consumers GC'd after 5s idle, no reconnect


namastex888:
## Summary

`AutomationEngine.start()` subscribes to NATS JetStream using **ephemeral** consumers (no `durable` flag). NATS server applies a default `inactive_threshold: 5s` to ephemeral consumers, so any trigger event type with an idle gap longer than 5 seconds causes the server-side consumer to be garbage-collected. The Bun NATS client's `for await (const msg of consumer)` then exits silently — no error logged, no reconnect, no re-subscribe — and every subsequent `eventBus.publish(...)` for that trigger type is accepted by the stream but delivered to nobody.

**Impact in production:** every automation whose `trigger_event_type` is low-frequency (minutes or hours between publishes) silently stops firing forever after the first idle window of >5s. The stream keeps accepting events, the sweeper/publisher side keeps advancing state as if events were consumed, but no handler ever runs. No alert, no log, no recovery until the API process restarts.

This is a silent data loss / correctness bug for any long-interval automation (idle follow-ups, scheduled hand-offs, archival side effects). **Severity: P0** — failure mode is indistinguishable from healthy operation at the API layer.

## Root cause chain

1. **Automation engine uses ephemeral consumers.**
   `packages/core/src/automations/engine.ts:117`
   ```ts
   const subscription = await eventBus.subscribePattern(`${eventType}.>`, async (event) => {
     await this.handleEvent(event);
   });
   ```
   No `durable` passed in `SubscribeOptions`.

2. **`buildConsumerConfig` does not set `inactive_threshold`.**
   `packages/core/src/events/nats/consumer.ts:27`
   Without an explicit `inactive_threshold`, NATS server assigns the default for ephemerals (5s in nats-server — confirmed on disk meta.inf: `\"inactive_threshold\":5000000000`). Consumers created via `generateConsumerName()` end up as pull consumers with 5s GC.

3. **`createSubscription` has no reconnect on iterator exit.**
   `packages/core/src/events/nats/subscription.ts` (the `processMessages` loop): when the server deletes the consumer, `for await (const msg of consumer)` returns cleanly (no exception) and the function simply returns. `isActive` is still true but nothing is driving it, and no code re-invokes `subscribeInternal`.

4. **The error path in the connection status watcher is also swallowed.**
[...]

PR #587, opened 2026-05-01
fix(automations): reconcile NATS subscriptions on toggle and dead-iterator (#546)

### LINKED ISSUE #546: automation engine: NATS consumer subscription becomes stale after automation enable→disable→enable cycle (requires omni-api restart to recover)
## Summary

Toggling an automation `enabled=true → false → true` via the API leaves the automation engine in a state where it's no longer consuming NATS events for that subject, even though the automation reads `enabled=true`. The durable consumer exists, the API state looks correct, the health endpoint returns `nats.connected=true`, but `chat.idle_timeout` events are not processed and no actions fire. **Only `pm2 restart omni-api` recovers.**

We hit this twice in 24h on a follow-up automation in production. Below are reproducible repros from each occurrence.

## Environment

- `@automagik/omni` v2.260424.3 (durable-consumer fix from #445 + handoff/sweeper fixes from #528 are deployed)
- Single-node deploy (omni-api + omni-nats both on the same host via pm2)
- Postgres 16, NATS JetStream embedded, Bun runtime
- Automation: `Eugenia follow-up (Haiku)` UUID `2c8c013b-a6a4-4a04-8c54-d6f6b3bf3952`, trigger `chat.idle_timeout`, action `call_agent` + `send_message`

## Symptom

After the toggle cycle:

- `GET /api/v2/automations/<id>` returns `enabled=true` ✅
- `GET /api/v2/health` returns `status=healthy`, `nats.connected=true` ✅
- `ls jetstream/$G/streams/SYSTEM/obs/automation-engine-chat-idle_timeout/` shows the durable consumer dir present ✅
- `o.dat` mtime frozen at the time of the disable/reset ❌
- Zero `chat.idle_timeout` events processed → no `call_agent` POSTs to the configured agent → no `send_message` actions
- `chat_follow_up_state` rows accumulate in `disarm_reason IS NULL` with `next_fire_at` passing without anyone reading

## Repro 1 — NATS reset without restarting omni-api

**Date**: 2026-04-27 11:13 BRT

Operation: drain stale NATS backlog by moving the durable consumer dir aside, then reset just the NATS server.

```bash
mv /home/genie/.omni/data/nats/jetstream/\$G/streams/SYSTEM/obs/automation-engine-chat-idle_timeout \
   /tmp/durable-consumer-removed-\$(date +%s)
pm2 restart omni-nats   # NATS server reset, fresh

# omni-api was NOT restarted (uptime: 2 days)
# Automation toggled to enabled=true via PATCH

# Result: 47 minutes of zero follow-ups firing
# - GET /api/v2/automations: enabled=true
# - GET /api/v2/health: nats.connected=true (TCP up)
# - But: ls obs/ — no automation-engine-chat-idle_timeout (consumer never recreated)
# - Agno /agents/eugenia-followup/runs received zero POSTs
```

Resolution: `pm2 restart omni-api`. First POST to agent fired 1m48s after restart.

## Repro 2 — automation disable→enable cycle via API

**Date**: 2026-04-27 15:00 BRT (~3h after Repro 1 was already fixed)

Operation: regression detector flipped `enabled=false` via PATCH; we re-enabled 8min later via PATCH `enabled=true`.

```bash
# 15:00:30 BRT — kill switch:
curl -X PATCH -d '{\"enabled\":false}' http://localhost:8882/api/v2/automations/<uuid>

# 15:08:?? BRT — manual re-enable:
curl -X PATCH -d '{\"enabled\":true}' http://localhost:8882/api/v2/automations/<uuid>

# Result: 9+ hours later, still zero follow-ups firing
# - GET /api/v2/automations: enabled=true
# - obs/automation-engine-chat-idle_timeout/o.dat mtime frozen at 14:55 BRT (5min before the disable)
# - omni-api uptime continuous (12h+, never restarted)
```

296 `chat_follow_up_state` rows accumulated in armed state during this 9h window with next_fire_at scheduled across that window.

Resolution (planned): `pm2 restart omni-api`.

## Suspected root cause

The automation engine binds the NATS subscription on startup and keeps an in-memory handle. The `enabled` flag in the automation config is checked on event delivery to gate action firing — but the subscription itself is not torn down/re-bound when the flag flips. After:

- A NATS server reset (Repro 1) — the server-side consumer/binding is gone, but the client handle thinks it's still subscribed and silently no-ops.
- A disable→enable toggle (Repro 2) — the handle remains, but something about the disable path leaves it in a state where event callbacks no longer fire after re-enable. (It's also possible the disable path `unsubscribe()`s but the re-enable path does not `subscribe()` — symmetric inverse missing.)

In both cases, `pm2 restart omni-api` reconstructs the engine bootstrap which calls `subscribe()` afresh and recovers.

## Why this is operationally bad

- **Silent failure** — no error in any health endpoint or log; just zero events processed.
- **Detection requires comparing `o.dat` mtime against expected event rate** or counting agent POSTs at the receiving end. We only caught it because we were specifically looking.
- **Worst case is worse than disabled** — operators see `enabled=true` and assume coverage; meanwhile the automation hasn't fired in hours.
- **Production-impacting toggles need a restart afterward**, which isn't in any runbook because the API contract suggests toggle is sufficient.
- A circuit-breaker / kill-switch pattern that disables and later re-enables an automation programmatically is unsafe today — re-enable doesn't actually re-enable.

## Proposed fix

In the automation engine module:

1. On `automation.enabled` PATCH transition `false → true`, call `subscribe(triggerEventType, durableName)` explicitly (idempotent — server returns existing consumer or creates new).
2. On `automation.enabled` PATCH transition `true → false`, call `unsubscribe()` and clear in-memory handle.
3. Add a periodic reconciler (e.g., every 60s) that compares each enabled automation's expected subscription against the engine's active subscriptions; re-subscribe if missing.
4. Ideally surface a healthcheck assertion: `automations_with_active_subscription_count == automations_enabled_count`. Currently health says \"NATS connected\" which is true at the TCP layer but lies about the application-layer binding.

## Workaround until fixed

After any of:

- NATS server restart
- Moving/deleting consumer state on disk
- PATCH to `automation.enabled`

…run `pm2 restart omni-api` and verify with `ls obs/automation-engine-<eventType>/` plus a positive count of recent target-agent POSTs.

## Cross-reference

- Related: #445 (closed) — ephemeral consumer GC after 5s; fixed by switching to durable. This issue is about the durable consumer's *binding* lifecycle on the omni-api side, not the consumer's existence on the NATS server side.
- Related: #528 (closed) — handoff/sweeper race; fixed in v2.260424.3.

vasconceloscezar:
## Summary

Fixes #546.

The automation engine kept its NATS subscriptions in a flat array and rebuilt them naively via `stop()` + `start()` inside `reload()`. After a disable→enable PATCH cycle (or a NATS server reset) the durable consumer's iterator could end up dead while the engine still considered itself subscribed. The API reported `enabled=true` and health reported `nats.connected=true`, yet no `chat.idle_timeout` events were processed and `pm2 restart omni-api` was the only known recovery — hit twice in 24h in production on a follow-up automation.

### Leak point

`packages/core/src/automations/engine.ts` (pre-fix `start()` / `reload()` / `stop()` around the old `subscriptions: Subscription[]` array, lines 85, 116–136, 151–164, 177–182):

- `reload()` called `stop()` then `start()`, which delegated subscription recreation to `eventBus.subscribePattern` against an already-existing durable. The bus path goes through `ensureConsumer` → `updateOrRecreateConsumer`. When this happens twice on the same durable across a toggle, the new client iterator can land in a state where it never receives deliveries — and the engine had no way to detect or recover from that.
- `Subscription` carried no liveness signal, so a dead iterator looked identical to a healthy one to the engine. The only \"recovery\" was the next process restart re-running `start()` from scratch.

### Fix

- Track subscriptions in a `Map<triggerEventType, Subscription>` instead of a flat array.
- Add a reconcile loop that:
  - drops subscriptions for triggers with no enabled automations,
  - re-subscribes when a subscription's iterator reports `!isAlive()`,
  - is idempotent on healthy state (no leaked handlers).
- Extend `Subscription` with an optional `isAlive()` probe and expose the NATS subscription wrapper's existing `isActive` flag through it (`packages/core/src/events/nats/subscription.ts`).
- Run the reconciler on `reload()` and on a 30s timer (`reconcileIntervalMs`, configurable, set to `0` in tests). Timer is `unref()`'d so it never blocks process shutdown.
- `reload()` now reconciles in place rather than tearing down and rebuilding, so a healthy subscription survives reloads where the trigger set hasn't changed (no churn on the durable consumer).

### Test coverage

`packages/core/src/automations/__tests__/engine-reconcile.test.ts`:

- disable→enable re-subscribes the durable consumer
- reconcile re-subscribes when the iterator dies
- reconcile is idempotent when subscriptions are healthy
- reload drops orphan-trigger subscriptions
- reload preserves a healthy subscription instead of churning it
- reload from never-started engine seeds automations without subscribing

The existing `engine-durable.test.ts` (#445 regression) still passes — durable name, queue, `startFrom=new`, and dedup-by-trigger semantics are unchanged.

## Test plan

- [x] \`bun test packages/core/src/automations/\` — 279 pass, 0 fail
- [x] \`bun test packages/core/src/events\` — 136 pass, 0 fail
- [x] \`bun test packages/api/src\` — 967 pass, 0 fail
- [x] \`bun test\` (full suite) — 3704 pass, 0 fail
- [x] \`make typecheck\` — clean
- [x] \`make lint\` — clean

### DIFF
--- a/packages/core/src/automations/__tests__/engine-reconcile.test.ts
+++ b/packages/core/src/automations/__tests__/engine-reconcile.test.ts
@@ -1,0 +1,186 @@
+/**
+ * Regression test for #546 — automation engine NATS subscription becomes
+ * stale after a disable→enable cycle (or NATS server reset), leaving the
+ * automation marked enabled in the API but silently not consuming events.
+ *
+ * Verifies that the engine:
+ *  - Re-subscribes after a disable→enable toggle (reload path).
+ *  - Re-subscribes when the underlying iterator dies (reconciler path).
+ *  - Does not leak duplicate subscriptions on no-op reconciles.
+ *  - Cleans up subscriptions when the last automation for a trigger is
+ *    disabled.
+ */
+
+import { afterEach, beforeEach, describe, expect, mock, test } from 'bun:test';
+import type { EventBus, GenericEventHandler, SubscribeOptions, Subscription } from '../../events/bus';
+import { createAutomationEngine } from '../engine';
+import type { Automation } from '../types';
+
+function makeAutomation(triggerEventType: string, id = 'auto-1', enabled = true): Automation {
+  return {
+    id,
+    name: `Auto ${id}`,
+    enabled,
+    priority: 0,
+    triggerEventType,
+    triggerConditions: [],
+    conditionLogic: 'and',
+    actions: [],
+    debounce: { mode: 'none' },
+  } as unknown as Automation;
+}
+
+interface FakeSubscription extends Subscription {
+  alive: boolean;
+}
+
+interface BusHarness {
+  bus: EventBus;
+  subscribeCalls: Array<{ pattern: string; options: SubscribeOptions }>;
+  liveSubs: FakeSubscription[];
+}
+
+function makeBus(): BusHarness {
+  const subscribeCalls: Array<{ pattern: string; options: SubscribeOptions }> = [];
+  const liveSubs: FakeSubscription[] = [];
+
+  const bus = {
+    connect: mock(async () => {}),
+    publish: mock(async () => ({ id: '', sequence: 0, stream: '' })),
+    publishGeneric: mock(async () => ({ id: '', sequence: 0, stream: '' })),
+    subscribe: mock(async () => ({ id: 'x', pattern: '*', unsubscribe: async () => {} }) as Subscription),
+    subscribePattern: mock(async (pattern: string, _handler: GenericEventHandler, options?: SubscribeOptions) => {
+      subscribeCalls.push({ pattern, options: options ?? {} });
+      const sub: FakeSubscription = {
+        id: `sub-${subscribeCalls.length}`,
+        pattern,
+        alive: true,
+        isAlive() {
+          return this.alive;
+        },
+        async unsubscribe() {
+          this.alive = false;
+        },
+      };
+      liveSubs.push(sub);
+      return sub;
+    }),
+    subscribeMany: mock(async () => ({ id: 'x', pattern: '*', unsubscribe: async () => {} }) as Subscription),
+    subscribeAll: mock(async () => ({ id: 'x', pattern: '*', unsubscribe: async () => {} }) as Subscription),
+    close: mock(async () => {}),
+    isConnected: mock(() => true),
+  } as unknown as EventBus;
+
+  return { bus, subscribeCalls, liveSubs };
+}
+
+describe('AutomationEngine — reconcile + toggle (#546)', () => {
+  let engine: ReturnType<typeof createAutomationEngine>;
+  let harness: BusHarness;
+
+  beforeEach(() => {
+    // reconcileIntervalMs=0 disables the periodic timer so tests drive
+    // reconcile() explicitly.
+    engine = createAutomationEngine({ defaultConcurrency: 1, reconcileIntervalMs: 0 });
+    harness = makeBus();
+  });
+
+  afterEach(async () => {
+    await engine.stop();
+  });
+
+  test('disable→enable toggle re-subscribes the durable consumer', async () => {
+    const auto = makeAutomation('chat.idle_timeout', 'a1', true);
+
+    await engine.start(harness.bus, [auto]);
+    expect(harness.subscribeCalls).toHaveLength(1);
+    expect(harness.subscribeCalls[0]?.options.durable).toBe('automation-engine-chat-idle_timeout');
+
+    // Simulate PATCH enabled=false: reload with empty enabled set.
+    await engine.reload([{ ...auto, enabled: false } as Automation]);
+
+    // The previous subscription must be unsubscribed so we don't leak
+    // handlers.
+    expect(harness.liveSubs[0]?.alive).toBe(false);
+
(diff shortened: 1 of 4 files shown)


vasconceloscezar:
@gemini-code-assist addressed in 941fd8cd:

- **L195 (high)** — `reconcileSubscriptions` now single-flight via a `reconcilePromise` slot. Concurrent callers (timer tick, manual `reconcile()`, `reload()`) await the in-flight pass instead of mutating the subscriptions Map in parallel; the slot clears in `finally` so failures don't deadlock the next call.
- **L285 (medium)** — `setInterval` replaced with a recursive `setTimeout` (`scheduleNextReconcile`). Next tick is only scheduled after the current pass resolves, with a `reconcileEnabled` flag so `stopReconcileTimer` cleanly halts the loop.

Two new regression tests in `engine-reconcile.test.ts`:
- 5 concurrent `reconcile()` against a dead sub → 1 re-subscribe (folds into single in-flight pass).
- `reconcile()` racing `reload()` against a dead sub → 1 alive sub, no leaked handles.

Suite: 8 pass / 0 fail (was 6).

Issue #542, opened 2026-04-27
follow-up disarm: handoff doesn't override prior non-terminal disarm reasons


namastex888:
## Summary

Production regression observed despite #528 fix: a lead got a follow-up message **15 min 56 sec after a handoff** because the chat_follow_up_state row was already disarmed with `customer_replied` 1 minute BEFORE the handoff fired. The synchronous disarm in `/send/handoff` (#535) is a no-op in this case, so the row stays in `customer_replied` instead of moving to `handoff`. Then `armForOutbound` does not block re-arming (customer_replied is not in TERMINAL_DISARM_REASONS), and a follow-up fires.

## Reproduction (production case 5594992438493)

Timeline (BRT):
- 15:44 lead inbound, 15:45 seller responses → row armed
- (sequence advances naturally)
- 17:37:17 lead inbound \"Já sou cliente\" → row disarmed reason=`customer_replied`
- 17:38:02 agent calls `human_handoff` → POST `/api/v2/messages/send/handoff`
  - `chats.update({settings: {agentPaused:true}})` succeeds
  - `services.followUpLifecycle.disarm({reason: 'handoff'})` is called but is a **no-op** because `disarmActive()` SQL: `WHERE disarmReason IS NULL` — the row is already disarmed
- 17:38:23 dispatcher logs \"Agent response suppressed — handoff triggered during run\" but `parts:1` was already sent (one chunk made it through plugin.sendMessage)
- 17:38 plugin emits `message.sent` for that chunk with senderAgentId=SELLER
- 17:38 follow-up-hooks `armForOutbound` runs:
  - reads existing row: disarmReason=`customer_replied`, disarmedAt=17:37
  - terminal-disarm guard checks `TERMINAL_DISARM_REASONS.has(disarmReason)` — `customer_replied` is NOT in the set, so guard does NOT trigger
  - upsertArmed re-arms the row: sequenceIndex=0, nextFireAt≈17:58
- ~17:53:58 a follow-up POST fires, agent generates: \"Mario, vou te conectar com nosso suporte técnico...\"

Database verification:
```sql
-- Final state of chat_follow_up_state row:
SELECT disarm_reason, disarmed_at, last_agent_message_at
FROM chat_follow_up_state cfs JOIN chats c ON c.id=cfs.chat_id
WHERE c.external_id='5594992438493';
-- disarm_reason | disarmed_at        | last_agent_message_at
[...]

PR #588, opened 2026-05-01
fix(api): handoff disarm overrides non-terminal prior reason (#542)

vasconceloscezar:
## Summary

- Fixes the Mario leak (production case 5594992438493 on 2026-04-26): a follow-up message fired ~16 min after a handoff because the `chat_follow_up_state` row was already disarmed with `customer_replied` 1 min before the handoff. The synchronous disarm shipped in #535 became a no-op, the row stayed in `customer_replied` (not in `TERMINAL_DISARM_REASONS`), and `armForOutbound` re-armed the sequence on a tail `message.sent`.
- Special-cases `reason === 'handoff'` in `disarmActive()` (`packages/api/src/services/follow-up-lifecycle.ts`): drop the `disarmReason IS NULL` guard, but still refuse to downgrade strong existing reasons (`HANDOFF_OVERRIDE_PROTECTED` = `TERMINAL_DISARM_REASONS` ∪ `contact_closed`). All other reasons keep the original idempotent semantics.
- After the fix, `disarm_reason` advances to `handoff` on top of `customer_replied` / `sequence_complete`, the terminal-disarm guard in `armForOutbound` triggers, and re-arm is refused.

## Test plan
- [x] New: customer_replied → handoff override (row advances to `handoff`)
- [x] New: sequence_complete → handoff override
- [x] New: session_cleared / contact_closed NOT downgraded by handoff
- [x] New: NULL → handoff fresh disarm (existing path still works)
- [x] New: handoff → handoff idempotent
- [x] New: full production scenario — customer_replied → handoff disarm → tail `armForOutbound` refused
- [x] `bun test packages/api/src/__tests__/follow-up-lifecycle.test.ts` — 26 pass
- [x] Full pre-push gate (`make typecheck` + `bun test`) — 3698 pass / 0 fail

Refs: #528, #535, #538, #539
Closes: #542

### DIFF
--- a/packages/api/src/__tests__/follow-up-lifecycle.test.ts
+++ b/packages/api/src/__tests__/follow-up-lifecycle.test.ts
@@ -256,9 +256,9 @@ describeWithDb('FollowUpLifecycleService (integration)', () => {
     expect(rows).toHaveLength(0);
   });
 
-  test('second disarm on an already-disarmed row is idempotent (no new event)', async () => {
-    // Non-handoff disarms remain strictly idempotent. A handoff override
-    // (#542) is exercised by the dedicated tests below.
+  test('second non-terminal disarm on an already-disarmed row is idempotent (no new event)', async () => {
+    // Non-terminal disarms remain strictly idempotent. Terminal overrides
+    // (#542) are exercised by the dedicated tests below.
     await service.armForOutbound({
       chatId: testChatId,
       instanceId: testInstanceId,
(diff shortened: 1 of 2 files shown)


vasconceloscezar:
@gemini-code-assist addressed in 47d1d733:

- **L80** — `TERMINAL_OVERRIDE_PROTECTED` now a `FollowUpDisarmReason[]` constant; the spread is module-load-once, not per disarm call. Renamed from `HANDOFF_OVERRIDE_PROTECTED` and the doc reflects the generalized intent.
- **L442** — guard switched from `reason === 'handoff'` to `TERMINAL_DISARM_REASONS.has(reason)`. Every terminal reason (`handoff`, `session_cleared`, `archived`, `window_expired`) now overrides non-terminal prior reasons; none can downgrade another terminal-or-stronger row.
- **L277** — flipped: the idempotency test now uses `sequence_complete` (non-terminal) on `customer_replied` so the no-override semantics still gets coverage. Added explicit overrides for `archived`, `session_cleared`, `window_expired`, plus an `archived` non-downgrade-of-contact_closed test.

Lifecycle suite: 30 pass / 0 fail (was 26).

PR #592, opened 2026-05-01
fix(api): sweep — resolve JIDs to UUID on remaining /chats/:id* POST/DELETE/PATCH routes

namastex888:
Follow-up to **#591** which patched the GET handlers (`/:id`, `/:id/messages`, `/:id/participants`, `/:id` PATCH/DELETE) — the live 500s in production. This sweep applies the same `resolveChatIdParam(c, raw, instanceId?)` resolution to every remaining `:id*` route in `routes/v2/chats.ts`. Each one had the same latent bug: passing `c.req.param('id')` straight into a service method that called `eq(<uuid_col>, raw)`.

## Routes covered

| Route | Body has `instanceId`? | Resolution path |
|---|---|---|
| `POST /:id/archive` | optional | passed to resolver |
| `POST /:id/unarchive` | optional | passed to resolver |
| `POST /:id/hide` | no body | active-instance fallback |
| `POST /:id/unhide` | no body | active-instance fallback |
| `POST /:id/label` | no | active-instance fallback |
| `DELETE /:id/label` | no | active-instance fallback |
| `POST /:id/pin` | required | passed to resolver |
| `POST /:id/unpin` | required | passed to resolver |
| `POST /:id/mute` | required | passed to resolver |
| `POST /:id/unmute` | required | passed to resolver |
| `POST /:id/participants` | no | active-instance fallback |
| `DELETE /:id/participants/:platformUserId` | no body | active-instance fallback |
| `PATCH /:id/participants/:platformUserId/role` | no | active-instance fallback |
| `POST /:id/read` | required | passed to resolver |
| `POST /:id/disappearing` | required | passed to resolver |
| `POST /:id/reopen-contact` | no body | active-instance fallback |

After this sweep, **every** `chatsRoutes.<verb>('/:id*')` handler in the file goes through `resolveChatIdParam` before touching a service. Routes that already accept an explicit `instanceId` in their body pass it to the resolver — no API-key context lookup needed, the explicit instanceId is authoritative.

## Behaviour change for unresolvable IDs

Pre-sweep: **500 Server error** with `PostgresError: invalid input syntax for type uuid: "120363...@g.us"`.

Post-sweep: **404** with the actionable message already pinned in #591:

```json
{
  "error": {
    "code": "NOT_FOUND",
    "message": "Chat not found: <raw>. Pass either a chat UUID or set the API key's active instance via POST /api/v2/context/use first."
  }
}
```

## Test plan

New describe block in `chats-id-jid-resolution.test.ts` — `POST /chats/:id/* — JID/external-id resolution sweep` (6 tests, 11 expects):

- [x] `POST /:id/read` with JID + no active instance → 404 (not 500)
- [x] `POST /:id/archive` uses `body.instanceId` for resolution (no active-context lookup needed)
- [x] `POST /:id/pin` with required `instanceId` in body resolves via that instance
- [x] `POST /:id/hide` (no body) falls back to active-instance context
- [x] `DELETE /:id/participants/:platformUserId` resolves the chat id from active instance
- [x] `PATCH /:id/participants/:platformUserId/role` resolves before the role update

```
bun test packages/api/src/routes/v2/__tests__/chats-id-jid-resolution.test.ts
 17 pass / 0 fail / 45 expect calls

bun test packages/api
 995 pass / 285 skip / 0 fail
typecheck clean (turbo full pipeline)
biome clean
```

## Why a separate PR (not folded into #591)

#591 was scoped to the 500 the operator surfaced in production logs. The sweep is mechanical (same one-line wrapper applied to 16 routes) but benefits from a clean diff that reviewers can audit at a glance — every hunk is the same shape. After this lands, the file has a uniform "resolve before service" invariant.

## Diff summary

```
packages/api/src/routes/v2/__tests__/chats-id-jid-resolution.test.ts | +104
packages/api/src/routes/v2/chats.ts                                  | +48 -16
2 files changed, 152 insertions(+), 16 deletions(-)
```

Pairs with #591.

### DIFF
--- a/packages/api/src/routes/v2/__tests__/chats-id-jid-resolution.test.ts
+++ b/packages/api/src/routes/v2/__tests__/chats-id-jid-resolution.test.ts
@@ -46,6 +46,12 @@ interface HarnessOptions {
   apiKeyRow?: { activeInstanceId: string | null; contextInstanceId: string | null } | null;
   /** When set, `findByExternalIdSmart` returns this chat (id-only). null = not found. */
   resolvedChat?: { id: string } | null;
+  /**
+   * When set, the API key is restricted to these `instanceIds` (mirroring the
+   * `apiKeys.instanceIds` column). Default `null` = unrestricted. Used by the
+   * cross-instance oracle tests to flip the api key into a scoped state.
+   */
+  apiKeyInstanceIds?: string[] | null;
 }
 
 interface HarnessCalls {
@@ -102,7 +108,13 @@ function mountHarness(options: HarnessOptions = {}): {
         }),
       },
     } as never);
-    c.set('apiKey', { id: 'test-key', name: 'test', scopes: ['*'], instanceIds: null, expiresAt: null } as never);
+    c.set('apiKey', {
+      id: 'test-key',
+      name: 'test',
+      scopes: ['*'],
+      instanceIds: options.apiKeyInstanceIds ?? null,
+      expiresAt: null,
+    } as never);
     await next();
   });
   app.route('/chats', chatsRoutes);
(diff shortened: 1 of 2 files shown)


namastex888:
Thanks for the catch — fixed in `0d20...` (latest push).

Moved `checkInstanceAccess(apiKey, instanceId)` **inside** `resolveChatIdParam`, before `findByExternalIdSmart`. This closes the existence oracle by making the unauthorised path return the same `403 VALIDATION` for both "chat exists in forbidden instance" and "chat absent in forbidden instance" — neither case touches the DB.

Behaviour matrix:

| chat in instance | apiKey scoped to instance | response |
|---|---|---|
| exists | yes | 200 (resolved) |
| absent | yes | 404 |
| exists | **no** | **403 VALIDATION (no DB hit)** |
| absent | **no** | **403 VALIDATION (no DB hit)** |
| n/a (UUID input) | n/a | resolver bypassed; handler-level `checkInstanceAccess` remains authoritative |
| unrestricted key (`instanceIds === null`) | n/a | passthrough (no behaviour change for single-instance deployments) |

Also covers the drift case: if `apiKey.instanceIds` is tightened *after* `activeInstanceId` was set via `POST /context/use`, the resolver catches the stale active context too — defense in depth.

New test block — `cross-instance existence oracle` (5 tests, 11 expects), all the cases above pinned.

```
22/22 chats-id-jid-resolution.test.ts pass
1000/1000 api package tests pass
typecheck clean, biome clean
```

Will follow up with the same fix on the merged GET routes from #591 if you want it as a separate PR for cleaner audit, but since the resolver itself is the surface we're hardening here, both code paths benefit from this single change.

PR #633, opened 2026-05-12
feat(release): channel-aware publishing (mirror genie wish release-channel-dev)

namastex888:
## Summary

Ports genie [PR #2419 — release-channel-dev wish](https://github.com/automagik-dev/genie/pull/2419) verbatim to omni. After merge, omni publishes dev-branch and main-branch builds to distinct channels with their own `.well-known/<channel>.json` pointers — same shape as genie.

Felipe directive 2026-05-12 (Phase A of pgserve+omni release-pipeline unification): "extract from genie the publishing method for both --dev and stable, lets unify, no reason for calling channel beta next, same for omni, we are moving forward into publishing omni and pgserve in the exact same padronized way".

## Branch → channel mapping

| Trigger | Branch | Channel | Manifest |
|---|---|---|---|
| `workflow_run` CI success on `main` | main | `stable` | `.well-known/latest.json` (back-compat name) + advances `dev.json` |
| `pull_request` closed on `dev` | dev | `dev` | `.well-known/dev.json` only — `latest.json` untouched |
| `workflow_dispatch` (manual) | (any) | `dev` (default) | `.well-known/dev.json` only |

A stable release also advances `dev.json` (single-pipeline reality today). A dev release only advances `dev.json` so `latest.json` stays at the prior stable.

### `.github/workflows/version.yml`
- Add dispatch step after tag push: derives `CHANNEL` from triggering branch, then `gh workflow run release.yml --field version=<v> --field channel=<C>`. Required because GITHUB_TOKEN-pushed tags don't fire `push:tags` workflows (GitHub anti-recursion guard); workflow_dispatch is the documented exception.

### `.github/workflows/release.yml`
- Add `channel` workflow_dispatch input (choices: `stable / beta / canary / dev`, default `stable`)
- Change `with: channel: stable` → `with: channel: ${{ inputs.channel || 'stable' }}` when calling release-publish.yml

### `.github/workflows/release-publish.yml`
- Add `dev` to the workflow_dispatch channel choices
- Drop the `if: channel == 'stable'` gate on the `.well-known` writer — every non-draft publish writes the manifest(s) for its channel
- Replace single-file `latest.json` writer with case-statement emitting per-channel files; stable also writes dev.json (single-pipeline)
- Use staged-then-`git diff --cached` shape (fixes the bootstrap-when-file-never-existed footgun genie surfaced in #2412)

## Out of scope (separate follow-up PR)

- **Consumer-side install.sh** with `OMNI_CHANNEL=dev|stable` support — omni's existing `install-client.sh` does sparse-clone+build-from-source; a new install.sh that downloads signed tarballs is a separate landing.
- **CLI `--dev` / `--stable` flags** on `omni update` — separate consumer-side PR matching genie's update.ts diff.
- **Pgserve Phase B** — pgserve will mirror this PR but main-only + stable-only (no dev channel per Felipe directive); separate larger refactor since pgserve currently uses workflow_run chaining instead of orchestrator pattern.

## Cross-repo references

- Genie source PR: https://github.com/automagik-dev/genie/pull/2419
- Genie wish: `.genie/wishes/release-channel-dev/WISH.md` (in genie repo)
- Parent pgserve wish: https://github.com/namastexlabs/pgserve/pull/125 (v3-prerelease-trust-loop)

## Test plan

- [x] Workflow YAML files diff-reviewed against genie PR #2419 line-by-line
- [x] `bash scripts/check-fingerprint-pinning.sh` — still passes (4-channel pin guard unaffected)
- [ ] Post-merge: cut a dev tag (e.g. by merging any dev PR) → confirm `.well-known/dev.json` appears + GH release marked `prerelease: true`
- [ ] Post-merge: cut a main tag (rolling PR dev→main merge) → confirm `.well-known/latest.json` AND `dev.json` both advance + GH release is full release

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -64,6 +64,7 @@ on:
           - stable
           - beta
           - canary
+          - dev
       draft:
         description: 'Create as draft (require manual promotion)?'
         required: true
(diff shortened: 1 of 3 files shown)


namastex888:
@chatgpt-codex-connector confirmed — this is a real P1, fixed in commit `2f814bc6`.

The dev-channel dispatch path would have hit your exact scenario: dev tags carry no `-rc.` / `-beta.` suffix, so the `publish-npm` job's `if:` gate was a no-op for them. `npm publish --tag latest` would have raced version.yml's inline `npm publish --tag next` (same version, 409), and the `npm dist-tag add … latest` fallback then promoted the dev cut to `@latest`.

**Fix: drop `publish-npm` job entirely.** version.yml is the single source of truth for npm publishing (branch-derived dist-tag: `main → @latest`, `dev → @next`). release.yml now only orchestrates the signed-tarball chain (build → sign-attest → publish → release-notes) — matches genie's release.yml shape post-cutover. No `inputs.channel` gate needed because the duplicate is gone.

Diff: 1 file, +16 / −65.

PR #657, opened 2026-05-22
feat(a2a): add v1 discovery and task runtime

filipexyz:
## Summary
- Implements A2A v1 JSON-RPC methods, v1 role/part/task/event shapes, caller-scoped task persistence, task list/get/cancel/subscribe, and Agent Card v1 discovery.
- Adds authenticated Omni A2A registry endpoints plus SDK/CLI helpers for listing agent cards and sending messages.
- Keeps authentication on Omni API keys only. WorkOS was removed from the implementation and is intentionally not included.
- Adds a small CLI pgserve discovery timeout stabilization for the test suite.

## Validation
- bunx biome check .
- focused A2A/channel/provider tests
- bun run typecheck
- bun run test
- pre-push gates: typecheck + full test suite, 3845 pass / 0 fail
- local 8882 smoke: a2a list + a2a send --wait returned TASK_STATE_COMPLETED with artifacts

## Notes
- A2A registry is designed for a server-side dashboard: list agents through /api/v2/a2a/agents, fetch the card through /api/v2/a2a/agents/:agentId/card, then call the instance A2A endpoint with the same Omni API key bearer auth.


packages/channel-a2a/src/a2a-handler.ts
```diff
@@ -77,149 +180,456 @@ function jsonRpcError(id: string | number | null, code: number, message: string,
...
+    contextId,
+  };
+  const taskStore = getTaskStore(ctx);
+  const task = await taskStore.createTask({
+    instanceId: ctx.instanceId,
+    taskId,
+    contextId,
+    message: normalizedMessage,
+    metadata: { ...(sendParams.metadata ?? {}), callerKey: getCallerKey(ctx) },
+  });
```
vasconceloscezar: **H1 — Cross-tenant task hijack via client-supplied `taskId`**

`taskId` comes from `sendParams.taskId ?? message.taskId ?? generateCorrelationId('a2a')` (line 418), then `taskStore.createTask({ taskId, ... })` writes unconditionally — `A2ATaskStore.set` has no existence check (`task-store.ts:184`). Attacker keyA can submit `SendMessage { taskId: "<victim-task-from-keyB>" }` and the prior task is replaced: new `callerKey: keyA`, history reset, status flipped to `TASK_STATE_WORKING`. After overwrite, `canAccessTask` (which only guards reads) lets keyA do anything to "its" task; keyB's task is gone.

Fix options:
1. Before `createTask`, `getTask(instanceId, taskId)`. If a task exists with a different `callerKey`, return `A2A_TASK_NOT_FOUND` (don't leak existence) or `RPC_INVALID_PARAMS`.
2. If the existing task is owned by the same `callerKey`, treat it as a follow-up — append to `history` instead of replacing. This is what A2A v1 expects for repeat `SendMessage` against an existing taskId.

The current behavior is both a tenant-isolation gap and a spec deviation.

packages/channel-a2a/src/a2a-handler.ts
```diff
@@ -77,149 +180,456 @@ function jsonRpcError(id: string | number | null, code: number, message: string,
...
 }
 
+function getTaskStore(ctx: A2AHandlerContext): A2ATaskStore {
+  ctx.taskStore ??= new A2ATaskStore();
+  return ctx.taskStore;
+}
+
+function getCallerKey(ctx: A2AHandlerContext): string {
+  return ctx.callerKey ?? 'anonymous';
+}
```
vasconceloscezar, line 276: **H2 — `'anonymous'` callerKey collapses unauthenticated callers into one tenant**

When `x-omni-api-key-id` is absent, every caller gets `callerKey: 'anonymous'`. In production `authMiddleware` always sets the header on the `/a2a/:instanceId` route (app.ts:217-222), so today this is defense-in-depth. But any future path that reaches `handleA2ARequest` via `plugin.handleWebhook` (e.g., a webhook adapter or test wiring) skips that gate, and every anonymous caller shares one task namespace — they can read/cancel each other's tasks.

Suggested defense in depth: reject create/read/cancel/subscribe when `callerKey === 'anonymous'` and return `UNAUTHORIZED`. Pair with a regression test that calls `handleA2ARequest` without the header and asserts rejection.

packages/channel-a2a/src/a2a-handler.ts
```diff
@@ -77,149 +180,456 @@ function jsonRpcError(id: string | number | null, code: number, message: string,
...
+  initialTask: A2ATask,
+): Promise<A2ATask> {
+  const deadline = Date.now() + syncWaitMs();
+  let current = initialTask;
+
+  while (Date.now() < deadline) {
+    const latest = await taskStore.getTask(instanceId, taskId);
+    if (latest) current = latest;
+    if (taskIsSettled(current)) return current;
+    await sleep(Math.min(SEND_WAIT_POLL_MS, Math.max(0, deadline - Date.now())));
```
vasconceloscezar: **M1 — 100 ms polling on every blocking SendMessage**

`waitForSettledTask` calls `taskStore.getTask()` every 100 ms for up to 30 s = 300 storage reads per blocking call. Once `PluginStorage` is backed by Postgres (per the file header), that's up to 300 DB round-trips per request, and ~10·N qps for N concurrent A2A clients just for polling.

Cheaper alternatives:
- Subscribe to the event bus for terminal task events keyed by `correlationId` and resolve on first event — single RTT in the happy path.
- Or at minimum, exponential backoff (100 → 250 → 500 → 1000 ms) — ~4× fewer round-trips with the same p95 latency.

packages/channel-a2a/src/stream-store.ts
```diff
@@ -78,15 +90,17 @@ export class A2AStreamStore {
...
+          rpcId,
         };
 
         entryRef = entry;
         store.streams.set(key, entry);
 
         // Auto-close on idle (no parts received)
         entry.closeTimer = setTimeout(() => {
-          store.closeStream(instanceId, taskId, 'failed');
+          store.closeStream(instanceId, taskId, 'TASK_STATE_FAILED');
```
vasconceloscezar: **M2 — Idle timeout marks long-running tasks as FAILED**

This fires `closeStream(..., 'TASK_STATE_FAILED')` after 30 s of stream idleness. Because `plugin.ts:35-37` wires `onClose` to `taskStore.updateStatus(..., state)`, the task itself flips to `TASK_STATE_FAILED` — even when the producer is still working and just slow to emit the first artifact. The status is then a lie about system state and downstream consumers (e.g., `GetTask`, `ListTasks { status: TASK_STATE_FAILED }`) see fake failures.

Two options:
1. Reset the idle timer from the dispatcher side whenever `taskStore.updateStatus('WORKING')` is called — keeps active tasks alive across slow segments.
2. On idle close, only close the SSE; do NOT propagate to task state. Let the actual completion/failure paths drive terminal state.

packages/api/src/services/a2a-discovery.ts
```diff
@@ -0,0 +1,216 @@
...
+  return skills;
+}
+
+function buildA2AAgentCard(params: {
+  baseUrl: string;
+  agent: Agent;
+  instance: Instance;
+  provider?: AgentProvider | null;
+  extended?: boolean;
+}): AgentCard {
```
vasconceloscezar, line 63: **M3 — `extended` is dead code**

Destructured here but never referenced in the function body. The routes (`a2a.ts:47`, `app.ts:171`) and `resolveA2AAgentCard` (`a2a-discovery.ts:184`) wire `extended` through three layers for no effect. Either implement the extended-card fields (richer `securitySchemes`, internal-only metadata, etc.) or drop the parameter from `buildA2AAgentCard`, `resolveA2AAgentCard`, and the call sites.

vasconceloscezar (commented):
# Review summary

Looked through the A2A v1 runtime, agent-card discovery, caller-scoped task store, SSE streams, and SDK helpers. Three things stand out as worth blocking on; the rest is polish.

## 🔴 High — fix before merge

**H1. Cross-tenant task hijack via client-supplied `taskId`** — see inline on `a2a-handler.ts`. `A2ATaskStore.createTask` overwrites unconditionally. A request with `{ taskId: <victim-task> }` replaces the victim's task; `canAccessTask` only guards reads, not creates. Also a spec deviation: per A2A v1, `SendMessage` with an existing `taskId` should append history, not replace.

**H2. `'anonymous'` callerKey collapses unauthenticated callers into one tenant** — see inline on `getCallerKey`. In production `authMiddleware` always sets `x-omni-api-key-id`, so this is defense-in-depth, but the handler should refuse to operate without a real callerKey rather than silently bucketing everyone into `'anonymous'`.

## 🟡 Medium

**M1. Blocking `SendMessage` polls the task store every 100 ms** for up to 30 s (`waitForSettledTask`). Once `PluginStorage` is Postgres-backed, that's up to 300 round-trips per blocking call, scaling to ~10·N qps for N concurrent clients. Either subscribe to the event bus for terminal task events, or back off the poll (100 → 250 → 500 → 1000 ms).

**M2. SSE idle timeout flips long-running tasks to FAILED.** `IDLE_CLOSE_MS = 30_000` fires `closeStream(..., 'TASK_STATE_FAILED')` which propagates to `taskStore.updateStatus(..., FAILED)` via `onClose`. A task whose first artifact takes >30 s is silently marked FAILED even though the producer is still working. Reset the idle timer on `updateStatus('WORKING')` from the dispatcher side, or — on idle close — only close the SSE and leave the task state alone.

**M3. `extended` is a dead parameter** in `buildA2AAgentCard` — destructured at line 65 but never read. The routes wire `extended: true` through three layers for no effect. Either implement the extended fields or drop the param.

**M4. NATS enum imports replaced with string-literal type assertions** (`consumer.ts`, `streams.ts`). Compiles today because the `nats` package's enums happen to be string-valued. If they ever change shape, JetStream config silently breaks. Either keep the runtime imports or add a unit test pinning the wire-level values.

**M5. Caller-key test coverage is one-deep** — only `GetTask` denial is tested. Missing: `CancelTask`, `SubscribeToTask`, cross-key taskId reuse (H1), `ListTasks` filtering, and the anonymous-bucket case (H2).

## 🟢 Low / nits

- L1 `taskIsSettled` (a2a-handler.ts:285) maps legacy `'input-required'` but not legacy `'completed'/'failed'/'canceled'`. Inconsistent with `taskIsTerminal` in `task-store.ts:28`.
- L2 `listTasks` loads all keys for an instance into memory before sorting/paginating (task-store.ts:87–124). O(N) memory per request — document as known limitation.
- L3 `nextPageToken` is a byte offset over a list sorted by `lastModified`. Inserts mid-pagination shift the boundary → duplicates/skips.
- L4 `A2APartSchema.passthrough()` + the `normalizePart` fallback `{ data: part, mediaType: 'application/json' }` lets arbitrary client-supplied fields land in `task.history`. Consider tightening once shape stabilizes.
- L5 `pgserve-transport.ts:47` — `TCP_DISCOVERY_TIMEOUT_MS` lowered 5 s → 1 s globally for test-suite stability. On a loaded CI runner this may cause flakier discovery in the real startup path. Consider gating by `NODE_ENV === 'test'`.
- L6 `a2a-discovery.ts:198–201` — single-active-instance fallback leaks "there is exactly one active a2a agent" via 200 vs 404 from `/.well-known/agent-card.json`. Minor side-channel.
- L7 `process.env.A2A_SEND_WAIT_MS` parsed per request (`a2a-handler.ts:281`). Memoize at module load.
- L8 No `.env.example` documentation for `A2A_ENABLED` / `A2A_SEND_WAIT_MS`.

## What's good

- Clean separation: `task-store` / `stream-store` / `a2a-handler` / `plugin`.
- Sound JSON-RPC error mapping: `a2aErrorInfo` produces proper `google.rpc.ErrorInfo` payloads with `reason` / `domain` / `metadata`.
- Legacy v0.3 method aliasing via `LEGACY_METHODS` + `validateVersion` is a nice compatibility move.
- Push-notification config explicitly rejected rather than silently ignored.
- `MAX_STREAMS` / `MAX_STREAMS_PER_INSTANCE` caps on the stream store — good DoS hardening.
- Feature-flagged behind `A2A_ENABLED=true` with a clean 503 stub.

## Recommendation

**FIX-FIRST.** H1 and H2 are exploitable as written. M1, M2, M5 are good to bundle with the same fix-up commit. The lows are polish.

PR #660, opened 2026-05-24
fix(pgserve): canonical default port (5432) + match PG major when reading legacy embedded data

automagik-genie:
Two broadly-applicable bugs in the canonical-pgserve path (both reproduced on a real upgraded host).

## 1. `DEFAULT_PGSERVE_PORT` hardcoded 8432 → `omni doctor` false-FAIL
`DEFAULT_PGSERVE_PORT` was `8432` (legacy bun-bridge era). It feeds `resolvePgservePort()` → the `PGSERVE_PORT` runtime env that `omni doctor`'s `pgserve-reachable` check probes. After the phase-3 socket-singleton cutover the canonical backbone binds **5432** over the UDS, so doctor reported:

```
✗ pgserve-reachable    cannot connect to pgserve on :8432
```

on **every healthy canonical host**, even though omni-api connects fine over the socket. Default is now `CANONICAL_PG_PORT` (5432). Stored `8432` entries in `~/.omni/config.json` are still treated as a stale default and re-resolved UDS-first by `resolveDatabaseUrl`.

## 2. `findAutopgPostgresBinary()` ignored PG major → 17→18 migration silently dies
It returned the *last* autopg `postgres` binary under `~/.local/share/autopg/<ver>/...` regardless of server major. A PostgreSQL server refuses to open a data dir from a different major ("database files are incompatible with server"), so on a v3-only host (PG 18) reading a **legacy PG 17** embedded dir (`~/.omni/data/pgserve`) it handed back PG 18 and the temp postmaster died — the `embedded-data-orphaned: temp postmaster exited before ready` symptom, leaving data stranded.

Now selects the binary whose major **matches** the embedded dir's `PG_VERSION`, and returns a version-specific skip reason ("no PostgreSQL 17 reader … install a matching autopg/reader") instead of silently failing.

## Tests
- `runtime-env.test.ts`: `resolvePgservePort` falls back to canonical 5432 (not 8432); honors explicit `server.pgservePort`; ignores non-positive overrides.
- `embedded-canonical-migration.test.ts` (new): `readDataDirMajor` reads PG_VERSION (17/18), null on absent/non-numeric.
- `doctor.test.ts`: legacy TCP-fallback assertion updated to the canonical port.
- Full `packages/cli` suite: **492 pass / 0 fail**; `turbo typecheck` 21/21; biome clean.

### DIFF
--- a/packages/cli/src/__tests__/doctor.test.ts
+++ b/packages/cli/src/__tests__/doctor.test.ts
@@ -670,8 +670,10 @@ describe('runDoctor — --fix mode', () => {
     // Accept three valid shapes:
     //   - Plain UDS form: postgres@localhost/omni  (PR #645+ post-postgres.js fix)
     //   - Legacy UDS w/ libpq: postgres@localhost/omni?host=… (pre-#645)
-    //   - TCP fallback: postgres@<host>:8432/omni
-    expect(startCall?.env?.DATABASE_URL).toMatch(/postgres@(localhost\/omni|[^/]+:8432\/omni)/);
+    //   - TCP fallback: postgres@<host>:<DEFAULT_PGSERVE_PORT>/omni (now canonical 5432)
+    expect(startCall?.env?.DATABASE_URL).toMatch(
+      new RegExp(`postgres@(localhost/omni|[^/]+:${DEFAULT_PGSERVE_PORT}/omni)`),
+    );
     // Recheck sees the drift as fixed.
     const drift = report.checks.find((c) => c.id === 'pm2-env-drift');
     expect(drift?.level).toBe('OK');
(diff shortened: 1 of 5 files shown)


automagik-genie:
Thanks @gemini-code-assist — addressed in `db1382b`: `findAutopgPostgresBinary` no longer depends on `readdirSync` order. Candidates are sorted by their `~/.local/share/autopg/<version>` dir (numeric, newest-first) via `compareVersionDesc` (exported + unit-tested, e.g. `2.6.10 > 2.6.1`), so it deterministically picks the highest patch within the wanted major (newest overall when major is unknown).

Note: the **GitGuardian Secrets Scan** failure on the first run was an org-level `no more API calls available` quota error (not a finding in this diff — the changes are pure TS with no secrets); this push re-triggers it.

PR #661, opened 2026-05-25
fix(pgserve): seamless cross-major embedded→canonical migration (auto-fetch matching reader)

automagik-genie:
## Problem (hit on a real upgraded host)
A user whose omni data lived in a **pre-canonical embedded cluster of a different PG major** (legacy **PG17** data; canonical now **PG18**) had it **silently stranded** after upgrading: a PostgreSQL server can only open a data dir of its own major, so the embedded→canonical migration couldn't read the old cluster and just skipped. Result: omni boots healthy on canonical but `omni instances list` returns nothing — the instances/messages/chats sit unmigrated in `~/.omni/data/pgserve`. Telling the operator to hand-install a matching PG reader is **not seamless**. (Found by dogfooding the previous fix, which only *diagnosed* the gap.)

## Fix — seamless across majors **and** schema drift
- **`resolveReaderForMajor()`**: prefer an installed autopg binary of the embedded major; otherwise **auto-fetch** `@embedded-postgres/<platform>@<major>` (cached under `~/.omni/cache/`), with ICU/SSL soname symlink fixup + `LD_LIBRARY_PATH` on spawn. The operator never deals with PG versions.
- **`copyTable` → TEXT format over the column INTERSECTION** (was BINARY/all-columns, which assumed identical major + schema). Survives the 17↔18 wire gap and drift between an old embedded schema and the current canonical one. Tables copied are `src ∩ dst`; obsolete embedded-only tables skipped.
- **Preserve install-local auth** (`MIGRATION_SKIP_TABLES` = `media_content, api_keys, api_key_audit_logs`): copying the stale embedded `__primary__` key over the fresh install's key broke `omni` CLI auth ("Invalid API key") right after migrating; copying the audit logs violated their FK to the preserved keys.

## Validation
Ran end-to-end on a real **PG17→PG18 stranded host**: auto-fetched the PG17 reader, copied **32 tables** in ~3s — **instances (2), messages (52), chats (6), persons (12)** recovered, `omni instances list` works, CLI auth intact.
Tests: new cases for `MIGRATION_SKIP_TABLES` (auth preserved) + `embeddedPostgresPackage` mapping; `readDataDirMajor`/`compareVersionDesc` retained. **Full `packages/cli` suite: 495 pass / 0 fail**, `turbo typecheck` clean, biome clean.

### DIFF
--- a/packages/cli/src/lib/__tests__/embedded-canonical-migration.test.ts
+++ b/packages/cli/src/lib/__tests__/embedded-canonical-migration.test.ts
@@ -3,7 +3,12 @@ import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 
-import { compareVersionDesc, readDataDirMajor } from '../embedded-canonical-migration.js';
+import {
+  MIGRATION_SKIP_TABLES,
+  compareVersionDesc,
+  embeddedPostgresPackage,
+  readDataDirMajor,
+} from '../embedded-canonical-migration.js';
 
 // readDataDirMajor underpins the matching-major reader selection: the temp
 // postmaster MUST be the same PG major as the data dir's catalog, or it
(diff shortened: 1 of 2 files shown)


automagik-genie:
Thanks @gemini-code-assist — all three addressed in `4a81667`:
- **version sort**: `latestReaderVersion` now sorts matches with `compareVersionDesc` and takes `[0]` (handles 17.10 vs 17.2 instead of trusting npm's order).
- **platform loader var**: the temp postmaster now sets `DYLD_LIBRARY_PATH` on macOS, `PATH` on Windows, `LD_LIBRARY_PATH` on Linux for fetched readers.
- **quoted identifiers**: `public."<table>"` in the \copy / TRUNCATE / count queries (column names were already quoted) to tolerate reserved words / case-sensitivity.

Note on the two red checks: both were infra/transient, not this diff — Quality Gate failed at the runner's *"Database failed to start"* pgserve-bootstrap step, and Secrets Scan hit the org GitGuardian `no more API calls available` quota. This push re-triggers both.

PR #663, opened 2026-05-26
fix(history-sync): preserve original messageTimestamp instead of ingest time

raphaelmarra:
## Summary

WhatsApp history sync messages were being stored with `platform_timestamp = now()` (ingest time) instead of the original `messageTimestamp` from the Baileys event. This caused all historical messages replayed on reconnect to appear as if they were sent at the moment of reconnection, breaking timeline ordering and conversation history display.

### Root causes fixed

**Bug 1 — `extractPlatformTimestamp` fell back to ingest time when messageTimestamp was absent**

Previously the function returned `new Date(fallback)` (where `fallback = Date.now()` at call site) when `messageTimestamp` was missing or zero. During history sync, many messages lack a `messageTimestamp`, so they silently got the current timestamp. Fixed: function now returns `Date | null`; callers handle null explicitly.

**Bug 2 — Protobuf Long object reconstruction was broken**

Baileys emits `messageTimestamp` as a protobuf `uint64` serialized after `deepSanitize` as `{ low, high, unsigned }`. The old code used only `ts.low`, losing the `high` word. Correct reconstruction:
```typescript
const lo = (ts.low >>> 0);   // treat as unsigned int32
const hi = (ts.high ?? 0) >>> 0;
tsNum = hi * 0x100000000 + lo;
```
This also handles negative `low` values (signed int32 wrapping at `0x80000000`).

**Bug 3 — `ingestMode` from EventMetadata was never read**

`EventMetadata.ingestMode?: 'realtime' | 'history-sync'` is set correctly by the NATS publisher, but `setupMessagePersistence` never consumed it. History-sync messages were always stored with `source = 'realtime'` instead of `source = 'sync'`.

**Bug 4 — `channel-whatsapp` `getMessageTimestamp` could propagate wrong timestamps**

`getMessageTimestamp` didn't handle missing/zero timestamps explicitly, risking `new Date(NaN)` or `new Date(0)` propagation.

## Changes

- `packages/api/src/plugins/message-persistence.ts`
  - `extractPlatformTimestamp`: returns `Date | null` (was `Date`)
  - Long reconstruction: `(hi >>> 0) * 0x100000000 + (lo >>> 0)`
  - `MessageMetadata` gains `ingestMode?: 'realtime' | 'history-sync'`
  - `handleMessageReceived`: reads `ingestMode`, sets `source = 'sync'` for history-sync, skips messages without `messageTimestamp` in history-sync path (can't store null — schema is NOT NULL), falls back to event timestamp only for realtime path
  
- `packages/api/src/plugins/agent-dispatcher.ts`
  - Two callers of `extractPlatformTimestamp` updated for `Date | null` return type

- `packages/channel-whatsapp/src/plugin.ts`
  - `getMessageTimestamp`: returns `Date | null`, skips messages without valid timestamp instead of publishing with wrong time

- `packages/api/src/plugins/__tests__/message-persistence.test.ts` (new)
  - 11 unit tests covering: undefined/missing/zero/empty input → null; number seconds; string seconds; Long with zero high; Long with non-zero high (post-2038); Long with negative low (signed int32 at 0x80000000); already-millisecond timestamp; no fallback contamination

## Reproduction

```sql
-- Before fix: all history-sync messages show platform_timestamp ≈ created_at (ingest time)
SELECT
  COUNT(*) FILTER (WHERE ABS(EXTRACT(EPOCH FROM (platform_timestamp - created_at))) < 5) AS wrong_timestamp,
  COUNT(*) FILTER (WHERE ABS(EXTRACT(EPOCH FROM (platform_timestamp - created_at))) >= 5) AS correct_timestamp
FROM messages
WHERE source = 'sync';
-- Before: wrong_timestamp ≈ 20297, correct_timestamp ≈ 0
-- After:  wrong_timestamp ≈ 0,     correct_timestamp ≈ N
```

Messages without `messageTimestamp` in history-sync are now skipped (not persisted) since the schema requires `platform_timestamp NOT NULL` and there is no recoverable correct timestamp for them.

## Test plan

- [x] All 11 unit tests for `extractPlatformTimestamp` pass (`bun test packages/api/src/plugins/__tests__/message-persistence.test.ts`)
- [x] TypeScript compilation passes (`bun run typecheck`)
- [x] Biome lint passes (`bun run lint`)
- [ ] Manual: reconnect a WhatsApp instance and verify `messages.platform_timestamp` matches `raw_payload->>'messageTimestamp'` for synced messages
- [ ] Manual: verify `messages.source = 'sync'` for history-sync messages

## Summary by CodeRabbit

* **Tests**
  * Added test suite validating timestamp extraction across various formats and edge cases.

* **Bug Fixes**
  * Hardened timestamp handling with fallbacks to prevent crashes when platform timestamps are missing or invalid.
  * History sync messages lacking valid timestamps are now skipped during ingestion.

* **Refactor**
  * Updated timestamp extraction logic to distinguish between realtime and history-sync message processing modes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/automagik-dev/omni/pull/663?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/api/src/plugins/__tests__/message-persistence.test.ts
+++ b/packages/api/src/plugins/__tests__/message-persistence.test.ts
@@ -1,0 +1,92 @@
+import { describe, expect, test } from 'bun:test';
+import { extractPlatformTimestamp } from '../message-persistence';
+
+describe('extractPlatformTimestamp', () => {
+  const FALLBACK = Date.now();
+
+  test('returns null when rawPayload is undefined', () => {
+    expect(extractPlatformTimestamp(undefined, FALLBACK)).toBeNull();
+  });
+
+  test('returns null when messageTimestamp is missing', () => {
+    expect(extractPlatformTimestamp({}, FALLBACK)).toBeNull();
+  });
+
+  test('returns null when messageTimestamp is 0', () => {
+    expect(extractPlatformTimestamp({ messageTimestamp: 0 }, FALLBACK)).toBeNull();
+  });
+
+  test('returns null when messageTimestamp is empty string', () => {
+    expect(extractPlatformTimestamp({ messageTimestamp: '' }, FALLBACK)).toBeNull();
+  });
+
+  test('handles number timestamp in seconds', () => {
+    const ts = 1746665253; // 2025-05-08
+    const result = extractPlatformTimestamp({ messageTimestamp: ts }, FALLBACK);
+    expect(result).not.toBeNull();
+    expect(result!.getTime()).toBe(ts * 1000);
+  });
+
+  test('handles string timestamp in seconds', () => {
+    const ts = '1746665253';
+    const result = extractPlatformTimestamp({ messageTimestamp: ts }, FALLBACK);
+    expect(result).not.toBeNull();
+    expect(result!.getTime()).toBe(1746665253 * 1000);
+  });
+
+  test('handles Long object (protobuf uint64) with non-zero low, zero high', () => {
+    // Baileys emits messageTimestamp as { low, high, unsigned } after deepSanitize
+    const ts = 1746665253; // 2025-05-08
+    const result = extractPlatformTimestamp(
+      { messageTimestamp: { low: ts, high: 0, unsigned: false } },
+      FALLBACK,
+    );
+    expect(result).not.toBeNull();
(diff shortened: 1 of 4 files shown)


namastex888:
@codex review this pr

PR #667, opened 2026-05-28
feat(follow-up): pre-fire human-active probe + 'human_active' disarm reason

lorenzomatheo:
## Problem

When a human operator takes over a chat directly in the channel inbox (e.g. Gupshup Agent Assist) **without** going through the `human_handoff` tool, none of the existing disarm paths fire:

- `chat.handoff_activated` is only emitted on the `agentPaused` `false → true` transition (set by `/api/v2/messages/send/handoff`). Manual takeover never flips it.
- `customer_replied` only disarms when the customer sends a new inbound message. The customer is happily talking to the human in between sweeper ticks.

Result: the follow-up sweeper nudges the customer with a generic "are you still interested?" right on top of the human-led conversation.

Production incident: `chat 616959139` on 2026-05-26 — operator started replying at 10:23 BRT, follow-up fired at **10:33:44 BRT**, 22 seconds after the human's `"Tá legal 😊"`. Full incident snapshot at [`brain/snapshots/cego-historico-gupshup-2026-05-28`](https://github.com/namastexlabs/genie-hapvida/blob/docs/eugenia-cards-1-2-3-2026-05-28/brain/snapshots/cego-historico-gupshup-2026-05-28/INVESTIGATION.md) (Caso B).

## Solution

Add a **pre-fire human-active probe** to the sweeper. Mirrors the existing `MessagingWindowProbe` pattern: optional, no-op when omitted, runs before `chat.idle_timeout` is published. Returning `'active'` disarms the row with a new `human_active` reason.

The probe is **DB-only**: it reads `chats.lastMessageAt` + `chats.lastMessageFromMe` that the sweep query already joins. Detection rule: outbound message strictly past `lastAgentMessageAt + graceMs` (default 2 min) ⇒ human took over. Inbound messages return `'inactive'` (the existing `customer_replied` path covers them). Missing columns return `'unknown'` (graceful degrade, preserves current behaviour).

No HTTP roundtrips, no new channel-plugin contract, no schema migration beyond a single enum value addition.

### Core (`packages/core/`)

- `events/types.ts`, `schemas/follow-up.ts`, `db/schema.ts` (`followUpDisarmReasons`): new `'human_active'` value on `FollowUpDisarmReason`. **Non-terminal** — re-arm is allowed on the next genuine agent reply, same semantics as `customer_replied`.
- `automations/follow-up/sweeper.ts`:
  - New `HumanActiveProbe` interface (async, mirrors `MessagingWindowProbe`).
  - `FollowUpStateRow` gains two optional fields: `chatLastMessageAt`, `chatLastMessageFromMe`.
  - `processRow` consults the probe between the window guard and the `chat.idle_timeout` emission; `'active'` → `disarm('human_active')`.
- `automations/follow-up/capabilities.ts`: new `createDbHumanActiveProbe({ graceMs })` factory.

### API service (`packages/api/`)

- `services/follow-up-sweeper.ts`:
  - `findAndLockDue` selects `chats.lastMessageAt` and `chats.lastMessageFromMe` and propagates them into the row.
  - Constructor instantiates `createDbHumanActiveProbe()` and passes it to `sweepFollowUps()`.

### Tests

- `packages/core/src/automations/follow-up/__tests__/sweeper.test.ts`: 4 new cases (`active` disarms with `human_active`; `inactive` fires; `unknown` fires; window guard runs first when both probes would disarm).
- `packages/core/src/automations/follow-up/__tests__/capabilities.test.ts`: 5 new cases covering the DB probe (active past grace; inactive within grace; inactive on inbound; unknown on missing columns; configurable graceMs).

```
Follow-up suite : 75 → 84 pass (9 new)
API services    : 192 pass
Core typecheck  : clean
Lint            : clean (warnings only — pre-existing)
```

## Out of scope (and why)

- **Channel-specific HTTP probe (e.g. Gupshup Partner API `GET /chats/:id`)** — `chats.lastMessageAt` is updated by the channel webhook pipeline for every message regardless of direction, so by the time the sweeper picks a candidate row, the human operator's outbound message is already visible. No side trip needed. A channel that wants stronger guarantees can layer an HTTP probe on top later.
- **Per-instance opt-in flag** — the DB probe is cheap enough (no extra query) and the detection rule is conservative (only outbound, only after a grace window) that defaulting it on for all instances is safe. Flag can be added later if any instance reports false positives.
- **`agentPaused: true` write** — out of scope here. The disarm alone stops the follow-up; the wider question of "should we also pause the agent dispatcher for this chat?" is a separate decision that belongs in a follow-up PR if needed.

## Re-arm semantics

`'human_active'` is intentionally **NOT** in `TERMINAL_DISARM_REASONS` (`packages/api/src/services/follow-up-lifecycle.ts`). When the bot eventually replies again on the same chat (human hands back, chat returns to the bot via normal `message.received` → agent dispatcher), `armForOutbound` re-arms the sequence as usual. Same semantics as `customer_replied` re-arm.

## Test plan

- [x] `bun test packages/core/src/automations/follow-up/` — 84/84 pass
- [x] `bun test packages/api/src/services/__tests__/` — 192/192 pass
- [x] Core typecheck (`bunx tsc --noEmit -p packages/core`) — clean
- [x] **Smoke test against origin/main (omni v2.260430.12) — bug reproduced**. See `scripts/test-human-active-disarm.sh`. Pre-merge run: row stayed armed (`disarm_reason = NULL`) and `nextFireAt` advanced — sweeper fired despite the simulated human takeover. Post-merge expected: same script flips to `disarm_reason = 'human_active'`.
- [ ] After merge, run the same smoke against a HML/PROD Omni build that includes this PR to confirm the fix end-to-end.
- [ ] Confirm next-agent-reply re-arms (i.e. `disarmReason` returns to NULL on subsequent armed sequence).

### Smoke script (`scripts/test-human-active-disarm.sh`)

```bash
PGURL='postgresql://user:pwd@127.0.0.1:9432/omni' \
  SWEEP_WAIT_SECONDS=25 \
  ./scripts/test-human-active-disarm.sh
```

The script (a) picks a gupshup chat, (b) re-arms its follow-up row, (c) injects `chats.lastMessageAt = NOW, lastMessageFromMe = true` (exactly what `createDbHumanActiveProbe` looks for), (d) waits one sweeper tick, (e) checks `chat_follow_up_state.disarm_reason`. Optional `CHAT_ID=<uuid>` override.

## Notes for reviewers

- The `'human_active'` enum value is added to **three** places (`events/types.ts`, `schemas/follow-up.ts`, `db/schema.ts`) intentionally — the codebase keeps these in sync by hand for type/runtime/zod parity. Followed the existing pattern.
- Pre-existing typecheck errors in `packages/api/src/app.ts:151`, `routes/v2/agent-routes.ts:14`, `routes/v2/instances.ts:25` are unrelated to this PR — they fail on `origin/main` too (verified via `git stash` round-trip).
- The 2 NATS test failures (`consumer.test.ts`, `streams.test.ts` — `RetentionPolicy not found`) are pre-existing import errors against the local `nats@2.29.3` install.

## Release Notes

* **New Features**
  * Follow-up automation now detects when a human agent is actively handling a chat and intelligently prevents idle timeout messages, eliminating duplicate automated messages during live agent assistance.

[Review Change Stack](https://app.coderabbit.ai/change-stack/automagik-dev/omni/pull/667?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/api/src/services/follow-up-sweeper.ts
+++ b/packages/api/src/services/follow-up-sweeper.ts
@@ -19,8 +19,10 @@ import {
   type EventBus,
   type FollowUpStateRepo,
   type FollowUpStateRow,
+  type HumanActiveProbe,
   type Logger,
   type MessagingWindowProbe,
+  createDbHumanActiveProbe,
   createLogger,
   createMessagingWindowProbe,
   sweepFollowUps,
@@ -34,6 +36,7 @@ const log = createLogger('follow-up-sweeper');
 export class FollowUpSweeperService {
   private readonly repo: FollowUpStateRepo;
   private readonly messagingWindowProbe: MessagingWindowProbe;
+  private readonly humanActiveProbe: HumanActiveProbe;
 
   constructor(
     private db: Database,
@@ -56,6 +59,13 @@ export class FollowUpSweeperService {
       },
     };
     this.messagingWindowProbe = createMessagingWindowProbe();
+    // DB-only probe: relies on `chats.lastMessageAt` / `lastMessageFromMe`
+    // already loaded by `findAndLockDue`. Cheap, no network. Disarms with
+    // `human_active` when the chat's most recent message is outbound and
+    // post-dates the last agent reply by more than the grace window —
+    // strong signal that a human operator took over directly in the
+    // channel inbox without going through `human_handoff`.
+    this.humanActiveProbe = createDbHumanActiveProbe();
   }
 
   /**
@@ -71,6 +81,7 @@ export class FollowUpSweeperService {
       eventBus: this.eventBus,
       logger: this.logger,
       messagingWindowProbe: this.messagingWindowProbe,
+      humanActiveProbe: this.humanActiveProbe,
     });
   }
 
@@ -110,6 +121,8 @@ export class FollowUpSweeperService {
           lastInboundCustomerMessageAt: chatFollowUpState.lastInboundCustomerMessageAt,
           chatName: chats.name,
           channelType: instances.channel,
+          chatLastMessageAt: chats.lastMessageAt,
+          chatLastMessageFromMe: chats.lastMessageFromMe,
         })
         .from(chatFollowUpState)
         .leftJoin(chats, eq(chatFollowUpState.chatId, chats.id))
@@ -135,6 +148,8 @@ export class FollowUpSweeperService {
         lastAgentMessageAt: r.lastAgentMessageAt,
         lastInboundCustomerMessageAt: r.lastInboundCustomerMessageAt ?? null,
         channelType: (r.channelType ?? null) as ChannelType | null,
+        chatLastMessageAt: r.chatLastMessageAt ?? null,
+        chatLastMessageFromMe: r.chatLastMessageFromMe ?? null,
       }),
     );
   }
--- a/packages/core/src/automations/follow-up/__tests__/capabilities.test.ts
+++ b/packages/core/src/automations/follow-up/__tests__/capabilities.test.ts
@@ -18,6 +18,7 @@ import {
   MESSAGING_WINDOW_MS,
   channelHasMessagingWindow,
   channelSupportsTypingIndicator,
+  createDbHumanActiveProbe,
   createMessagingWindowProbe,
   playTypingIndicator,
 } from '../capabilities';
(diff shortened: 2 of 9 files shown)


lorenzomatheo:
## ⚠️ Converted to draft — DB probe doesn't cover the real Caso B

Found a critical gap in the original design while validating against the production incident this PR was supposed to fix. Reverting to draft until the implementation pivots.

### The problem

The DB-only probe (`createDbHumanActiveProbe`) reads `chats.lastMessageAt` + `chats.lastMessageFromMe`. Those columns get updated by the inbound webhook pipeline, but the gupshup channel handler at `packages/channel-gupshup/src/handlers/webhooks.ts:247` filters explicitly:

```ts
if (webhook.event_type !== 'user_input') {
  logger.debug('[gupshup] non-user_input event ignored', { instanceId, event_type: webhook.event_type });
  return;
}
```

`user_input` is the customer → business event. When the operator handles the chat directly inside Gupshup Agent Assist (the actual Caso B scenario in `brain/snapshots/cego-historico-gupshup-2026-05-28`), no `user_input` is generated — the operator's reply never updates `chats.lastMessageAt`/`lastMessageFromMe` from the Omni's perspective. The probe never sees the takeover.

I verified this empirically: the smoke script (`scripts/test-human-active-disarm.sh`) **simulated** the DB update via UPDATE statement and reproduced the disarm path. But that's a synthetic path that doesn't exist in production. The chat returns to the sweeper with stale `lastMessageAt` and the follow-up fires anyway.

### What's correct here and worth keeping

- `'human_active'` added to `FollowUpDisarmReason` everywhere (events/types.ts, schemas/follow-up.ts, db/schema.ts, lifecycle's `TERMINAL_DISARM_REASONS` correctly omits it).
- `HumanActiveProbe` interface on `SweeperDeps` — async, optional, no-op when omitted, mirror of `MessagingWindowProbe`. Solid contract.
- `processRow` hook calls the probe between the messaging-window guard and the `chat.idle_timeout` emission. Order is right (window guard wins, tests cover this).
- `FollowUpStateRow` gains `chatLastMessageAt` / `chatLastMessageFromMe` (optional fields the sweep query is already selecting).
- 9 new tests covering the disarm path, ordering vs window guard, the unknown/inactive fall-throughs, and the DB probe factory's branches.

This is reusable infra regardless of which probe implementation lands on top.

### What needs to be replaced

The `createDbHumanActiveProbe()` factory in `packages/core/src/automations/follow-up/capabilities.ts`. The detection signal needs to come from somewhere the operator's takeover actually reaches.

### The path forward (from the public Gupshup docs)

Gupshup Agent Assist already emits the right signal — they call it [**Assignment Event**](https://console-docs.gupshup.io/docs/webhook):

> Assignment Event: Fires whenever an assignment occurs in the system.

Configured in **Settings → Webhooks → Create Webhook**. The chat's [own status](https://console-docs.gupshup.io/docs/managing-chat-status-for-effective-workflow) goes to `'Open'` when an agent takes it, which is the canonical "human is actively handling this" signal.

Concrete plan (separate follow-up PR, not this one):

1. Extend `packages/channel-gupshup/src/handlers/webhooks.ts` to **also** process Assignment Event payloads (not just `user_input`). Schema needs confirmation with `[email protected]` or empirical capture once the subscription is live.
2. New column on `chats` (or a dedicated `chat_agent_assist_activity` row) that stores `lastAgentAssistActivityAt` + `assignedAgentId`. Updated by the new handler.
3. Replace `createDbHumanActiveProbe` with `createAgentAssistActivityProbe` that reads that column. Same `HumanActiveProbe` interface — drop-in.
4. Customer (Hapvida in our case) toggles the Assignment Event webhook on in their Gupshup console pointing at our Omni webhook URL. Documented in the integration guide.

### Why not HTTP probe (External Assignment API)

[`External Assignment`](https://console-docs.gupshup.io/docs/chat-management-assignment-rules-chat-rules-slas-re-open-time) exists but the API spec is private — only released on request via `[email protected]`. Plus PULL adds 1 GET per sweep candidate. PUSH via Assignment Event is cheaper (zero overhead per tick), documented officially, and uses infra we already have.

### Status

- Smoke script (`scripts/test-human-active-disarm.sh`) **kept** — it reproduces the synthetic-path success and remains useful for any future probe that does land. After the real implementation, the smoke needs a companion test that hits the webhook path (probably a fixture in `packages/channel-gupshup/src/__tests__/`).
- Will reopen as Ready-for-Review when (a) Assignment Event webhook payload is captured/documented, (b) the new column + handler land, (c) the probe is swapped, (d) end-to-end smoke passes against a real operator takeover in HML.

Sorry for the back-and-forth — the validation caught this before any of us merged something that wouldn't fix the production case.

PR #691, opened 2026-06-01
fix(whatsapp): resolve reaction UUID targets

namastex888:
## Summary
- Promote the WhatsApp reaction UUID target fix to `dev`
- Resolve internal Omni message UUIDs to provider-native external IDs before plugin send
- Fail closed for UUID targets when target chat/message scope cannot be verified
- Add regression coverage for UUID happy path, missing chat/message, cross-chat mismatch, and external-ID fallback

## Test Plan
- `bunx biome check packages/api/src/routes/v2/messages.ts packages/api/src/routes/v2/__tests__/messages-send-reaction.test.ts`
- `bun test packages/api/src/routes/v2/__tests__/messages-send-reaction.test.ts packages/cli/src/__tests__/react-context.test.ts`
- `bun run --filter @omni/api typecheck`
- `bun run --filter @automagik/omni build:server`
- pre-push full suite on the sibling `felipe` PR branch: `3963 pass / 292 skip / 0 fail`
- dogfood: reacted between Felipe's two WhatsApp instances using an Omni UUID; async reaction record indexed without fallback text duplication

### DIFF
--- a/packages/api/src/routes/v2/__tests__/messages-send-reaction.test.ts
+++ b/packages/api/src/routes/v2/__tests__/messages-send-reaction.test.ts
@@ -1,0 +1,208 @@
+import { describe, expect, mock, test } from 'bun:test';
+import { NotFoundError } from '@omni/core';
+import { Hono } from 'hono';
+import type { AppVariables } from '../../../types';
+import { messagesRoutes } from '../messages';
+
+const INSTANCE_ID = '11111111-1111-4111-8111-111111111111';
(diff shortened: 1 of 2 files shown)


namastex888:
Addressed the Gemini review feedback: narrowed UUID target lookup error handling so only `NotFoundError` is converted to the reaction-target 404, unexpected lookup/system errors now propagate, and tests now mock `NotFoundError` plus cover the non-masked unexpected-error path. Re-ran targeted gates and pre-push suite after force-pushing the amended commit.

PR #707, opened 2026-06-08
fix(channel-gupshup): accept simplified HV-Entry-Flow payload

moraisdev:
## Problem

A Gupshup integration can send a **simplified inbound webhook payload** (a "payload clean-up" variant) instead of the native shape:

```json
{ "sender": { "id": "...", "name": "..." },
  "message": { "text": "...", "timestamp": "..." },
  "event": { "project_id": "..." } }
```

This shape fails `GupshupNativeWebhookSchema` — which requires `sender` as a string plus `messageobj` / `senderobj` / `botname` / `channel` / `destination` / `event_type` — so the webhook is acked (200) but **dropped as `dropped_unrecognized_shape`**. The inbound message never reaches the agent dispatcher.

## Fix

- `GupshupSimplifiedWebhookSchema` for the simplified shape
- `normalizeSimplifiedWebhook()` maps it onto the native inbound shape (synthesizes a dedupe-stable message id when none is provided; `Math.floor` keeps the unix-seconds timestamp an integer; accepts seconds or milliseconds)
- The webhook handler falls back **native → simplified** before dropping (extracted `resolveInboundWebhook` to keep `handleGupshupWebhook` within the cognitive-complexity limit)
- Downstream consumers unchanged; the native payload keeps working

## Tests

- Unit tests for the schema + normalizer (fractional/ms timestamps, explicit `message.id`, and that it rejects both the native shape and unrelated routing envelopes)
- **End-to-end handler tests** proving the simplified payload dispatches an **identical** inbound message (same phone + text) as the native one, and is not dropped
- All `channel-gupshup` tests pass · biome + typecheck clean

### DIFF
--- a/packages/channel-gupshup/src/__tests__/webhooks.test.ts
+++ b/packages/channel-gupshup/src/__tests__/webhooks.test.ts
@@ -14,7 +14,7 @@
 import { describe, expect, it } from 'bun:test';
 import * as fs from 'node:fs';
 import * as path from 'node:path';
-import { handleGupshupWebhook } from '../handlers/webhooks';
+import { GupshupSimplifiedWebhookSchema, handleGupshupWebhook, parseSimplifiedWebhook } from '../handlers/webhooks';
 import { GUPSHUP_WEBHOOK_METRIC } from '../observability';
 import type { GupshupPlugin } from '../plugin';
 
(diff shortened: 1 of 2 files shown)


moraisdev:
Atualizado — comentário de review endereçado e conflito resolvido:

- **@gemini-code-assist** ✅ `toUnixSeconds` agora envolve o retorno em `Math.floor`, garantindo inteiro mesmo com timestamp fracionário (`messageobj.timestamp` exige `int`). Adicionei um teste dedicado (`floors a fractional timestamp to an integer`).
- **Conflito** ✅ rebaseado em cima do `dev` atual (a mudança agora fica em cima do trabalho de observability / `async_response` / KHAL session ids). Reaplicada como fallback **aditivo** — formatos nativo e simplificado ambos parseiam.
- Extraí o helper `resolveInboundWebhook` pra manter o `handleGupshupWebhook` abaixo do limite de complexidade cognitiva do biome.

**Validação:** 39 testes do `channel-gupshup` passam · biome limpo · typecheck limpo. (As 2 falhas da suíte completa — `pinManifestEntry readonly-directory` e um CLI basic test — são pré-existentes/de ambiente, sem relação com esta mudança.)

PR #736, opened 2026-06-24
fix: handle private Slack media downloads

filipexyz:
## Summary

Fixes private Slack media downloads so stored media is real media bytes instead of Slack login HTML.

## Root Cause

Slack private file URLs can redirect from `files.slack.com` to `files-pri.slack.com`. The default fetch redirect behavior does not preserve the Authorization header across that cross-origin redirect, so protected media downloads could return an HTML login page instead of the expected image/audio/document bytes.

The `/messages/media/download` response also rebuilt the media serving URL with a duplicated instance id, which made the local cached media URL return 404.

## Changes

- Preserve Slack bot token authorization across Slack-owned private file redirects.
- Reject HTML responses when a non-HTML media type is expected.
- Allow `/messages/media/download` to resolve messages by `{ instanceId, chatExternalId, externalId }` for consumers that only have platform message references.
- Return the correct `/api/v2/media/{mediaLocalPath}` URL after caching.
- Add focused regression coverage for Slack private redirects and HTML rejection.

## Validation

- `bun test packages/api/src/services/__tests__/media-storage.test.ts packages/channel-slack/src/__tests__/files.test.ts`
- `bun run typecheck`
- `bun run build`
- pre-push full test suite: `4035 pass`, `294 skip`, `0 fail`

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -9,7 +9,7 @@
     {
       "name": "omni",
       "description": "Full Omni platform control — multichannel messaging, automations, events, batch ops via the omni CLI",
-      "version": "2.260622.1",
+      "version": "2.260624.3",
       "author": {
         "name": "Automagik"
       },
--- a/apps/ui/package.json
+++ b/apps/ui/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@omni/ui",
-  "version": "2.260622.1",
+  "version": "2.260624.3",
   "private": true,
   "type": "module",
   "scripts": {
--- a/bun.lock
+++ b/bun.lock
@@ -18,7 +18,7 @@
     },
     "apps/ui": {
       "name": "@omni/ui",
-      "version": "2.260622.1",
+      "version": "2.260624.3",
       "dependencies": {
         "@omni/sdk": "workspace:*",
         "@radix-ui/react-dialog": "^1.1.15",
@@ -51,7 +51,7 @@
     },
     "packages/api": {
       "name": "@omni/api",
-      "version": "2.260622.1",
+      "version": "2.260624.3",
       "dependencies": {
         "@google/genai": "^1.0.0",
         "@hono/swagger-ui": "^0.4.1",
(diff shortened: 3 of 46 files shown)


vasconceloscezar:
## Review (after merging `dev` + resolving conflicts)

Merged `origin/dev` into the branch and resolved two conflicts — both were the same whatsapp `processHistoryMessage` complexity refactor done independently on `dev` (`isHistoryMessageWithinSyncRange`) and here (`shouldSkipHistoryMessageByTimestamp`). The two helpers were byte-for-byte equivalent (inverted return), so I kept `dev`'s canonical one and dropped the duplicate. Also took `dev`'s `as unknown as string` over `as any` in the gupshup test. Pre-push suite green (4365 tests).

### Verdict: SHIP-ready, one cleanup worth a follow-up

**Correct & secure** ✅
- The auth-across-redirect fix is the right call: `fetch` strips `Authorization` on cross-origin redirects, and `files.slack.com → files-pri.slack.com` shares the `slack.com` suffix. Crucially the token is preserved **only when both the current and next hop match the allowed suffix** — so a redirect that leaves slack.com drops the token. No credential leak. 👍
- HTML-sniff rejection (content-type `text/html` OR body `<!doctype html`/`<html`) correctly catches the Slack login page returned with HTTP 200, and the "plain text mentioning <html>" test guards against false positives.
- The duplicated-instanceId URL fix (`/api/v2/media/${instanceId}/${mediaLocalPath}` → `/api/v2/media/${mediaLocalPath}`) is a real 404 fix — `mediaLocalPath` already contains the instance id.
- Tests assert the token rides across the 302 (both fetch calls carry it). Solid coverage.

**Cleanup (non-blocking, follow-up):**
- `normalizeMimeType`, `isHtmlMime`, `looksLikeHtml`, `hostMatchesSuffix` are **duplicated** verbatim in `packages/api/src/services/media-storage.ts` and `packages/channel-slack/src/handlers/files.ts`, and the manual-redirect-preserve-auth loop exists in both (`fetchWithOptionalAuthenticatedRedirects` vs `fetchSlackPrivateUrl`). Per the repo's reuse guidance these belong in a shared util (e.g. `@omni/core`) so the HTML-sniff + redirect-auth policy lives in one place. Worth a follow-up PR.

**Nit:** in `/media/download`, `services.instances.getById(instanceId)` now runs just before `checkInstanceAccess` — consider swapping the order so access is checked first (no leak today since the token is only used after the check, but cleaner).

PR #741, opened 2026-06-24
feat(agent-dispatcher): human handoff on agent errors for Gupshup

vasconceloscezar:
On Gupshup instances, any customer-facing agent error now routes the chat to a **human** instead of sending a dead-end error reply.

## Behavior
When an error would otherwise show the user a generic error message, on a native-handoff channel we instead:
1. Send **"Tô com um probleminha técnico aqui agora. Logo alguém do time vai entrar em contato com você."** AS the native Gupshup `HANDOFF` payload — one shot: delivers the text to the user **and** signals the human queue.
2. Set `chats.settings.agentPaused = true` (agent stops replying).
3. Disarm any active follow-up sequence (`reason: 'handoff'`).
4. Write a `handoff_logs` audit row.

This mirrors exactly what `POST /messages/send/handoff` does — just triggered automatically on error.

## Covers both error surfaces (per the ask)
- **Dispatch failure after retries** — the old English "Sorry, I ran into an issue…" path (`sendErrorFeedback`).
- **Provider-leaked error** blocked by the customer-safe guard — `agno-provider` now sets `result.metadata.customerErrorBlocked`, and the dispatcher reads it. (Gupshup is `canStreamResponse: false`, so this only flows through non-stream `trigger()` — no streaming plumbing needed.)

## Scope / safety
- Gated on `plugin.capabilities.canHandoff === true` — **Gupshup-only today**, but generalizes to any future native-handoff channel without code changes.
- **Non-handoff channels are unchanged** — they still get the plain error message.
- `triggerErrorHandoff` is resilient: on a delivery failure it returns `false` and the caller falls back to the normal error message.
- Message overridable via `OMNI_AGENT_ERROR_HANDOFF_MESSAGE`.

## Design note
I kept the route (`POST /messages/send/handoff`) **untouched** — it has no test coverage and uses a different plugin-access path (`channelRegistry`) than the dispatcher (`getPlugin`). `triggerErrorHandoff` is a dispatcher-local mirror of the route's side-effects, with a comment cross-referencing it. A future refactor could extract a shared `executeHandoff` once the route gets test coverage.

## Tests
- `resolveErrorHandoffMessage` precedence (env override / blank fallthrough).
- `agno-provider` sets `customerErrorBlocked: true` on a blocked error and `false` on a normal response.
- Full pre-push suite green.

## Not covered
The reaction-trigger send path (emoji reactions) doesn't apply the error-handoff — reactions producing a customer-facing provider error is an edge case; left as a follow-up if needed.

### DIFF
--- a/packages/api/src/plugins/__tests__/agent-dispatch-error-handoff.test.ts
+++ b/packages/api/src/plugins/__tests__/agent-dispatch-error-handoff.test.ts
@@ -12,10 +12,10 @@ import { describe, expect, it, mock } from 'bun:test';
 // Configurable plugin mock for `getPlugin`. Tests mutate `pluginState` per case.
 const pluginState: {
   canHandoff: boolean;
-  sendMessage: (...args: unknown[]) => Promise<{ messageId: string }>;
+  sendMessage: (...args: unknown[]) => Promise<{ success: boolean; messageId?: string; error?: string }>;
 } = {
   canHandoff: true,
-  sendMessage: mock(() => Promise.resolve({ messageId: 'msg-1' })),
+  sendMessage: mock(() => Promise.resolve({ success: true, messageId: 'msg-1' })),
 };
 
 mock.module('../loader', () => ({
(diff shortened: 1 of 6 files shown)


vasconceloscezar:
Addressed the HIGH-priority finding from @gemini-code-assist (commit a18487f4).

**Double-message bug — fixed.** `triggerErrorHandoff` now splits the two failure modes:
- **Delivery failure** (`plugin.sendMessage` throws) → returns `false`. The user got nothing, so the caller's plain-error fallback is correct (no double message).
- **Side-effect failure** (`chats.update` / `disarm` / `handoffLogs` throws after a successful send) → caught, logged as `agent_dispatch_error_handoff_side_effects_failed`, and **still returns `true`**. The user already received the handoff message, so no second message is sent.

Added a regression test (`agent-dispatch-error-handoff.test.ts`) covering all four branches: non-handoff channel, delivery failure, side-effect failure (→ true), and happy path.

On the `toPhone` type guard: `chatId` is typed `string` in the function signature, so the `typeof chatId === 'string' ? chatId : ''` guard would be dead code here — skipped to avoid noise. If a non-string ever reaches this path it's a type violation upstream, not a runtime input.

PR #761, opened 2026-07-03
feat: remote media mode — S3/MinIO presigned URLs

namastex888:
## Remote media mode — S3/MinIO presigned URLs

Opt-in `OMNI_MEDIA_MODE=remote`: store WhatsApp/Telegram media in **S3/MinIO** and hand agents a **presigned download URL** (via the existing `ProviderFile.url`) instead of a local filesystem path. **Local mode is byte-for-byte unchanged**; the k8s Helm chart defaults it **on**.

### Why
Splitting omni into k8s with a remote agent broke media: omni handed agents *local file paths* (`nats-genie-provider` → `ProviderFile.path`), which only works when omni + agent share a disk. Remote mode decouples them — the agent fetches media over HTTP from a presigned URL — and keeps omni-api **stateless** (media in object storage, not a PVC), so the HPA multi-replica story holds.

### What changed (5 groups, each independently SHIP-reviewed)
- **G1** — pluggable media storage backend (`local` | `s3`) behind `MediaStorageService`, using Bun-native `S3Client` (put + presign). **Zero new deps.** Local unchanged; remote records the stable **S3 key** on the row.
- **G2** — relocated the backend to `@omni/channel-sdk` (avoids an api→channel import cycle); rerouted **Telegram** (`store`) and **WhatsApp** (`storeStream`, size-guard preserved) ingest to the backend; presign minted **at dispatch** → `ProviderFile.url` + in-text URL. Audio stays URL-less (transcribed).
- **G3** — MinIO StatefulSet + idempotent bucket-bootstrap + omni-api S3 env in `deploy/helm/omni`; `values-dev` bundled+remote, `values-prod` external S3 via `existingSecret` (no plaintext creds); fail-guard for the misconfigured combo.
- **G5** — `media-processor` reads S3 bytes (temp file + `finally` cleanup) so remote-mode **audio transcription + image vision** work (closed the GA blocker).
- **G4** — `.env.example`, `docs/_internal/remote-media-mode.md`, and an e2e test (remote store → presigned URL → GET returns exact bytes).

### Evidence
- Whole-wish execution review: **SHIP** — 7/7 success criteria met, no scope creep, no local-mode regression, no plaintext creds, size-guard survives streaming, temp cleanup verified.
- Full suite green: **`bun test` → 4105 pass, 298 skip, 0 fail** (4403 tests); `typecheck` 21/21; `biome` clean.
- Integration tests run real `minio/minio` containers (shared singleton harness) — presigned GET returns bytes, expired presign rejected.

### Notes
- **Stacked on #759** (base `feat/deploy-k8s`) because G3 extends the omni Helm chart introduced there. Re-target to `dev` after #759 merges.
- Known pre-existing (not introduced here): a `baileys` ESM-mock leak that only surfaces under some combined runs; per-package suites pass and the full `bun test` above is 0-fail.

Wish: `.genie/wishes/omni-remote-media-mode/WISH.md`.


namastex888:
## Review-comment adjudication (fixed in 120a214d)

| Finding | Ruling | Action |
|---|---|---|
| **Codex P2** — remote mode persists dead `mediaUrl` (`messages.ts:622`) | **VALID — upgraded to HIGH.** `GET /api/v2/media` served via disk-only `readMedia`, so every remote-stored WhatsApp `mediaUrl` 404'd | Route now reads through the backend (`readMediaViaBackend` → S3 GET in remote mode, disk in local) + e2e case |
| **Gemini security-high** — path traversal in `LocalMediaBackend` | **VALID, downgraded to MEDIUM** — all current callers pass server-built/sanitized keys, and the HTTP route has its own per-segment guard; but the backend is an exported channel-sdk contract | `getSafePath` guard on `store`/`storeStream`/`read` + traversal tests |
| **Gemini medium** — temp leak if `writeFile` fails (`media-processor.ts:300`) | **VALID — LOW** (ENOSPC edge, self-limiting in tmpdir) | try/catch + `rm` on failure |
| **Gemini medium** — temp leak between materialize and `try` (`media-processor.ts:439`) | **INVALID** — only `log.info` sits in the window; `getPromptOverride` is evaluated inside the `try`. A structured log call throwing is not a practical failure mode | No change |

Gate after fixes: root `bun test` 4410 tests / 0 fail, typecheck 21/21, biome clean.

PR #808, opened 2026-07-10
feat(deploy): self-host backup tooling + data-safety knobs

namastex888:
## What

Make the **bundled self-host path** (`values-dev.yaml` — in-cluster autopg + MinIO + NATS) safe and ergonomic. **Zero changes to the AWS cloud realms.**

## Why

Running the self-host stack for real surfaced gaps that hit every self-hoster:
- **No backup story at all** — nothing dumps Postgres (WhatsApp/Baileys session state + messages) or mirrors the MinIO media.
- Footguns — `local-path` PV reclaim is `Delete`; rebuilding the fixed `:dev` image doesn't roll the pod; the dev omni-api memory limit (768Mi) is tight under media processing.

## Changes

- **`deploy/backup/`** — host-side `omni-backup.sh` (Postgres via a version-matched `postgres:18` client + MinIO `mc mirror`; **MinIO creds read from the cluster at runtime, no committed secrets**) and `omni-restore.sh` (verify a dump into a throwaway container, or restore into a live DB with a `--confirm` guard), plus a README with the traps, Time Machine note, launchd/systemd scheduling, and a restore runbook. Backups stay local; **no in-cluster CronJob**.
- **PVC retention knob** — `persistentVolumeClaimRetentionPolicy` on the bundled autopg/minio/nats StatefulSets, **rendered only when set**; `values-dev.yaml` opts into `Retain`.
- **`make redeploy`** — rebuild the image + `rollout restart` (a plain `deploy` won't roll the pod because the `:dev` tag is fixed with `pullPolicy: Never`).
- **omni-api memory** `768Mi → 1536Mi` in `values-dev.yaml` only.
- **docs** — `deploy/README.md` "Data safety & backups" (incl. the `local-path` PV `reclaim=Delete` note + the `Retain` one-liner); gitignore the per-machine `values-local.yaml` overlay.

## Cloud safety (the important bit)

The retention field renders **only when the value is set**, and only the self-host overlay sets it — so `helm template` for **prod / homolog / hml-alb is byte-identical to `origin/dev`** (verified with a before/after diff). Untouched: `values-{homolog,prod,hml-alb}.yaml`, `deploy/k8s/omni-hml/*`, `templates/deployment.yaml`, `_helpers.tpl`, `templates/ingress.yaml`, and `Chart.yaml`'s `appVersion`.

## Verification

- `helm lint` clean; `template REALM=dev` shows the 3 retention blocks + `memory: 1536Mi`.
- `make deploy-smoke` green — fresh k3d cluster, every pod Ready + `/health` 200.
- prod / homolog / hml-alb renders **diff-empty** vs `origin/dev`.
- `omni-backup.sh` + `omni-restore.sh` run end-to-end against a live install — restore verified into a throwaway `postgres:18` with row-count parity.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,6 @@ version.json
 
 # Helm dependency build artifacts (unpacked charts/ is source of truth)
 deploy/helm/omni/charts/*.tgz
+
+# Personal per-machine helm overrides (never commit)
+deploy/helm/omni/values-local.yaml
--- a/deploy/Makefile
+++ b/deploy/Makefile
@@ -66,6 +66,13 @@ deploy: ## Install/upgrade the release to ns $(NAMESPACE)
 	  --namespace $(NAMESPACE) --create-namespace \
 	  -f $(VALUES) $(HELM_EXTRA) --wait --timeout 5m
 
+.PHONY: redeploy
+redeploy: build ## Rebuild omni-api image + rollout restart (repick the fixed :dev tag)
+	# The dev image tag is fixed (:dev) with pullPolicy Never, so a plain `deploy`
+	# leaves the pod on the old image after a rebuild — force a rollout.
+	kubectl -n $(NAMESPACE) rollout restart deploy/$(RELEASE)
+	kubectl -n $(NAMESPACE) rollout status deploy/$(RELEASE) --timeout=5m
+
 .PHONY: status
 status: ## Rollout status + pods
 	kubectl -n $(NAMESPACE) rollout status deploy/$(RELEASE) --timeout=5m || true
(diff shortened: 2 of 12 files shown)


namastex888:
Addressed all review findings in `864d8418`:

**codex P1×2 (real bugs):** verify mode now propagates `pg_restore` failures (corrupt dump ⇒ verify FAILS); live restores use `--role` so the app role keeps ownership (new `RESTORE_ROLE` env, defaults to the target db).
**codex P2:** MinIO Service name is auto-derived from the release Secret (the chart names them identically), plus `PG_SVC`/`MINIO_SVC` overrides.
**gemini high:** portable base64 decode (GNU/BSD/busybox).
**gemini medium×6:** centralized EXIT cleanup for forwards/containers, `kill -0` fail-fast on port-forward bind failures, throwaway-postgres readiness asserted before `createdb`, `{1..30}`.

Bonus found while re-verifying: backgrounding the `kc()` wrapper leaked the real `kubectl` port-forward (subshell $! ≠ kubectl PID) — now backgrounds kubectl directly. Reproduced the leak live, re-ran backup + restore-verify e2e after the fix: both PASS, zero leaked forwards/containers.

PR #826, opened 2026-07-13
fix(api): discard superseded agent replies when newer inbound arrives mid-run

moraisdev:
## Problem

When a user sends a message while the agent is still processing the previous batch, the in-flight reply is generated from a stale snapshot. The user experience is:

1. User sends message A → debounce window closes → agent run starts with A only
2. User sends message B seconds later → it lands in the debouncer's in-flight buffer (by design)
3. The reply to A is delivered anyway — often **re-asking exactly what B just answered**
4. The finally-block re-flush dispatches B → a second reply lands moments later

The longer the agent run (tool calls, slow upstreams), the wider the race window — with a ~60s run, the reply to B arrives over a minute late and reads as the agent "answering out of nowhere". Measured on a production instance over 14 days, this two-reply signature appeared in ~23% of active chats (median gap between the two user messages: ~10s, i.e. beyond any reasonable debounce window — a longer window alone cannot fix this, and it taxes every turn with extra latency).

## Change

Opt-in, per-instance stale-reply policy — `messageSupersedeMode: 'off' | 'discard'` (default `'off'`, existing behavior unchanged):

- **`MessageDebouncer.hasPending(instanceId, chatId)`** — exposes whether newer inbound is buffered for the chat (accumulated while a flush is in flight).
- **Delivery gate** in `dispatchViaProvider` and `dispatchViaLegacy`, mirroring the existing handoff suppression: right before sending, if newer inbound is pending and the mode is `'discard'`, the stale reply is dropped and logged (`Agent response discarded — superseded by newer inbound`). The debouncer's finally-block re-flush then dispatches the buffered messages immediately, so the follow-up run answers everything with full context — the user sees **one coherent reply** instead of a stale one plus a correction.
- A reply that accompanies a handoff (`agentPaused` flipped during the run) is **never** discarded — that announcement must reach the user.
- `maybeTriggerErrorHandoff` extracted from `dispatchViaProvider` (keeps cognitive complexity within the lint budget).

### Why discard is safe

The provider has already persisted the turn on its side; discarding only skips **delivery**. The immediate re-flush run sees the full history (including the undelivered turn) plus the new message, so its reply is consistent with everything the user actually said. Conversation-wise the visible thread reads: A, B → one reply covering both.

## Config

```
PATCH /api/v2/instances/:id { "messageSupersedeMode": "discard" }
```

New column `instances.message_supersede_mode` (varchar(20), NOT NULL, default `'off'`) — migration `0039_instances_message_supersede_mode.sql` (hand-written ADD COLUMN, following the 0038 precedent to avoid the interactive rename prompts from `drizzle-kit generate`).

## Out of scope

- **Streaming dispatch**: delivery is progressive; there is no single pre-send point to gate. Documented in the commit.
- **Reaction triggers**: not routed through the message debouncer.
- Route-level override of the new setting (instance-level only for now; trivial follow-up if needed).

## Tests

- `message-debouncer.test.ts`: `hasPending` reports messages buffered during an in-flight flush and is scoped per `instanceId:chatId` (21 pass).
- Existing dispatcher suites unaffected (97 pass across `agent-dispatcher`, `agent-dispatch-error-handoff`, `agent-dispatcher-retry`).
- `turbo typecheck` green for `@omni/api`, `@omni/core`, `@omni/db`; `biome check` clean.

## Rollout

Ship with default `'off'`; enable per instance on a staging environment first, observe the `superseded by newer inbound` log line and reply coherence, then opt in production instances. If validation holds, a follow-up can discuss making `'discard'` the default, since delivering a knowingly-stale reply is arguably always wrong.

https://claude.ai/code/session_014uGa7nFuraP8ymBoVcBLTZ

### DIFF
--- a/packages/api/src/plugins/__tests__/message-debouncer.test.ts
+++ b/packages/api/src/plugins/__tests__/message-debouncer.test.ts
@@ -579,4 +579,49 @@ describe('MessageDebouncer', () => {
       expect(byKey['inst-2:chat-1']).toBe(1);
     });
   });
+
+  describe('hasPending — stale-reply detection', () => {
+    it('reports pending messages buffered while a flush is in flight', async () => {
+      const pendingObservedDuringFlush: boolean[] = [];
+      let release: (() => void) | undefined;
+      const gate = new Promise<void>((resolve) => {
+        release = resolve;
+      });
+
+      debouncer = new MessageDebouncer(async (_chatKey, messages) => {
+        if (messages[0]?.payload.content?.text === 'first') {
+          // Hold the flush open so a second message lands in the buffer.
+          await gate;
+          pendingObservedDuringFlush.push(debouncer.hasPending('inst-1', 'chat-1'));
+        }
+      });
+
+      debouncer.buffer('inst-1', 'chat-1', makeMessage('first'), disabledConfig);
+      await wait(20); // flush for 'first' is now in flight, awaiting the gate
+
+      expect(debouncer.hasPending('inst-1', 'chat-1')).toBe(false);
+      debouncer.buffer('inst-1', 'chat-1', makeMessage('second'), disabledConfig);
+      expect(debouncer.hasPending('inst-1', 'chat-1')).toBe(true);
+
+      release?.();
+      await wait(30); // in-flight flush completes + finally re-flush drains 'second'
+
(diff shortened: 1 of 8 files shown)


namastex888:
@moraisdev good job

PR #852, opened 2026-07-18
fix(api): enforce effective instance authority ceiling

automagik-genie:
## Security fix

Closes the effective instance-authority delegation gap reported on rolling production PR #839 and the partial-PATCH race found during review of this PR.

A key-management caller can now grant authority only when the child remains a subset under every enforcement interpretation currently present in the API:

1. canonical legacy `instanceIds` (`null` unrestricted, `[]` deny-all),
2. historical legacy routes where an empty list is treated as inactive,
3. profile-aware `instanceAllowlist` using `isLockActive`, and
4. routes where legacy and profile-aware restrictions intersect.

For authority-bearing PATCHes (`scopes` or `instanceIds`), `ApiKeyService.updateWithAuthorityGuard()` now:

- starts a PostgreSQL transaction;
- locks the target key row with `SELECT ... FOR UPDATE`;
- constructs the complete post-update record using persisted values for omitted fields;
- validates the complete scope and instance authority while the lock is held;
- updates the row in the same transaction.

This prevents:

- profile-only or legacy-only substitution;
- contradictory restrictions hiding broader authority behind an empty intersection;
- `instanceIds: []` acting as a hidden unrestricted grant on historical routes;
- scope-only PATCHes granting authority to a broader target;
- instance-only PATCHes activating retained scopes the caller does not hold;
- concurrent partial PATCHes composing from the same stale snapshot.

Metadata-only PATCHes retain the existing lightweight update path.

Additional hardening:

- malformed runtime key context fails closed;
- profile names are validated before lock derivation;
- UUIDs are canonicalized before set comparisons;
- primary-key rename protection is preserved inside the locked update path.

## Scope

Exactly four files:

- `packages/api/src/routes/v2/keys.ts`
- `packages/api/src/routes/v2/__tests__/keys-mint-ceiling.test.ts`
- `packages/api/src/services/api-keys.ts`
- `packages/api/src/__tests__/api-key-enforcement.test.ts`

No schema, migration, package, dependency, Helm, or workflow changes.

## TDD evidence

Observed RED before implementation:

- profile-only caller could mint broader authority;
- contradictory legacy/profile restrictions could bypass a single-intersection check;
- restricted callers could grant an empty legacy array consumed as unrestricted by historical routes;
- scope-only PATCH against a broader target returned `200` instead of `403`;
- instance-only PATCH activating retained scopes returned `200` instead of `403`;
- the atomic service method did not exist at the type boundary.

Final focused route result: **53 passed, 0 failed**.

## PostgreSQL concurrency proof

A disposable PostgreSQL 18 cluster was initialized locally, all **40/40 migrations** were applied, and the DB-backed API-key suite passed **17/17**.

The concurrency regression holds the first update's row lock, starts a conflicting second partial update, verifies the second guard does not run against stale state, releases the first transaction, then proves the second guard observes the first committed scopes and denies the conflicting update. The final row preserves only the first authorized change.

## Canonical verification

- `bun install --frozen-lockfile`: clean, no changes
- `bun test packages apps/ui`: **4,315 passed, 327 skipped, 0 failed** across 344 files
- `bun run typecheck`: **21/21**
- `bun run lint`: **1,364 files**, clean
- `bun run build`: **5/5**
- `bun run knip`: clean
- `bun run verify:versions`: **22/22**
- `git diff --check`: clean
- added-line static security scan: clean

Reviewed commit: `4459ffc252a86ea06ecbf8d56d2af63882a58866`  
Reviewed tree: `c82a362d2652dc20a0accf90e40b36fa34a4cd03`  
Committed diff SHA-256 against `origin/dev@1fd0061aba3759afcd5449b4a536d825ff7a4cd9`: `7f0115075ac7600ba9df047ed39f877b06b921faefc36bcb8b6df3d2373b1399`.

## Promotion safety

Merge to `dev` with a normal merge commit only after independent review and green CI. Then promote the exact settled dev tree through `homolog`; do not push or merge directly to `main`.

### DIFF
--- a/packages/api/src/__tests__/api-key-enforcement.test.ts
+++ b/packages/api/src/__tests__/api-key-enforcement.test.ts
@@ -246,6 +246,68 @@ describeWithDb('API Key Enforcement', () => {
   });
 
   // ============================================================================
+  // Atomic authority-bearing updates
+  // ============================================================================
+
+  describe('ApiKeyService.updateWithAuthorityGuard()', () => {
+    test('serializes concurrent partial authority updates on the same key', async () => {
+      const created = await apiKeyService.create({
+        name: `authority-race-${Date.now()}`,
+        scopes: ['keys:write'],
+        instanceIds: ['11111111-1111-4111-8111-111111111111'],
+      });
+      cleanupIds.keys.push(created.key.id);
+
+      let releaseFirst!: () => void;
+      const firstMayCommit = new Promise<void>((resolve) => {
+        releaseFirst = resolve;
+      });
+      let firstGuardEntered!: () => void;
+      const firstIsLocked = new Promise<void>((resolve) => {
+        firstGuardEntered = resolve;
+      });
+
+      const firstUpdate = apiKeyService.updateWithAuthorityGuard(
+        created.key.id,
+        { scopes: ['messages:send'] },
+        async () => {
+          firstGuardEntered();
+          await firstMayCommit;
+          return true;
+        },
+      );
+
+      await firstIsLocked;
+
+      let secondGuardScopes: string[] | undefined;
+      const secondUpdate = apiKeyService.updateWithAuthorityGuard(
+        created.key.id,
+        { instanceIds: ['22222222-2222-4222-8222-222222222222'] },
+        (_current, next) => {
+          secondGuardScopes = next.scopes;
+          return !next.scopes.includes('messages:send');
+        },
+      );
+
+      // The second guard must not run against the stale row while the first
+      // transaction holds its lock.
+      await new Promise((resolve) => setTimeout(resolve, 50));
+      expect(secondGuardScopes).toBeUndefined();
+
+      releaseFirst();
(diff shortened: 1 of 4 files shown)


automagik-genie:
@coderabbitai review

automagik-genie:
Implemented the complete partial-PATCH/TOCTOU repair in `4459ffc252a86ea06ecbf8d56d2af63882a58866` (tree `c82a362d2652dc20a0accf90e40b36fa34a4cd03`, exact reviewed diff SHA-256 `7f0115075ac7600ba9df047ed39f877b06b921faefc36bcb8b6df3d2373b1399`).

The authority-bearing PATCH path now locks the target row with `SELECT ... FOR UPDATE`, builds the complete post-update record from locked persisted state plus the request, enforces both scope and instance ceilings under that lock, and writes in the same transaction. This closes both the scope-only broader-target case and the inverse instance-only retained-scope case, including concurrent stale composition. A real migrated PostgreSQL concurrency regression passes.

@coderabbitai full review

PR #862, opened 2026-07-19
fix(api): bound key delegation allowlists

automagik-genie:
## Security repair

Closes the caller-authority gap reported during the refreshed homolog → main review (#839): profile-aware `chatAllowlist` and `outboundRecipientAllowlist` grants are now bounded to the caller just like scopes and instance access.

### Behavior

- Derives active/inactive authority with the same `isLockActive()` semantics used at runtime.
- Requires child chat and outbound-recipient authority to be a subset of the caller independently.
- Rejects restricted-caller escapes through:
  - another chat or outbound recipient;
  - a child profile whose corresponding lock is inactive;
  - legacy key creation with unrestricted profile allowlists;
  - metadata-only PATCH of an out-of-authority target.
- Applies to profile create, legacy create, and every PATCH using complete post-update state inside `updateWithAuthorityGuard`.
- Fails closed when caller profile/chat/outbound context is missing or malformed.
- Preserves unrestricted callers and in-authority subset behavior.

### TDD and verification

- RED: 7 isolated regressions failed on the pre-fix route behavior for cross-chat, inactive-chat, cross-recipient, missing caller arrays, and metadata-only out-of-authority PATCH paths.
- GREEN: `bun test packages/api/src/routes/v2/__tests__/keys-mint-ceiling.test.ts` — **68 passed, 0 failed**.
- Canonical suite: `bun test packages apps/ui` — **4,330 passed, 327 skipped, 0 failed** across 344 files.
- `bun run typecheck` — **21/21**.
- `bun run lint` — **1,364 files clean**.
- `bun run build` — pass.
- `bun run verify:versions` — **22/22**.
- `bunx knip` — pass.
- `git diff --check` — pass.
- Commitlint — pass.

### Frozen revision

- Base: `b4b64e0d5de1f7c2e529aade8cd07570fbaf5f0a`
- Head: `ccef271d06d5b018070fc30b8d279cccc9f9c070`
- Tree: `fbedbb0c4bcba808ac87f86cddb3b279eeb369bb`
- Commit diff SHA-256: `9107a582fda48350b6fbed834f5785e007ae278f2595d51dae73a49819b262c3`

This repair must settle on `dev`, then be promoted through an exact-tree two-parent merge to `homolog` before #839 is considered ready. Main remains unmerged pending Felipe's explicit decision.

### DIFF
--- a/packages/api/src/routes/v2/__tests__/keys-mint-ceiling.test.ts
+++ b/packages/api/src/routes/v2/__tests__/keys-mint-ceiling.test.ts
@@ -23,6 +23,10 @@
  *     routes that treat an empty legacy list as inactive. Unrestricted
  *     authority, contradictory restrictions, malformed context, and UUID case
  *     differences must not bypass any ceiling.
+ *   - Profile-aware `chatAllowlist` and `outboundRecipientAllowlist` grants are
+ *     independently bounded to the caller. A restricted caller cannot mint an
+ *     unrestricted/legacy child or mutate an out-of-authority target through a
+ *     metadata-only PATCH; malformed caller allowlist context fails closed.
  *
  * These tests mount the REAL `scopeEnforcerMiddleware` in front of the REAL
  * keys routes (the existing mint tests omit it — reviewer LOW-2), so the proof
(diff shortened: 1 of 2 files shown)


automagik-genie:
@coderabbitai review

PR #865, opened 2026-07-21
fix(api): sanitize media text before Postgres write (null bytes)

vasconceloscezar:
## Problem

Processed media content — transcription (Whisper), image/video description (vision), document extraction (pdf-parse) — is written straight to Postgres text columns by `MessageService`. When the extracted text contains a **NUL byte (0x00)** — common in PDF text extraction — the write fails:

```
invalid byte sequence for encoding "UTF8": 0x00
```

`awaitMediaProcessing` then reads the column state as `error`, and the agent-dispatcher hands the agent **`[media processing unavailable]`** instead of the real content. The agent silently never sees the document/image (it replies "I can't view images / media unavailable"), even though vision/extraction succeeded.

The codebase already ships `sanitizeText()` (utils/utf8.ts — it strips NUL bytes because PostgreSQL text columns reject them), but the four media-update methods do not use it.

## Fix

Apply `sanitizeText()` in `updateTranscription`, `updateImageDescription`, `updateVideoDescription`, `updateDocumentExtraction` before the `.set(...)`.

## Verification (built + tested locally, both images)

`PATCH /messages/:id/transcription` with a JSON NUL escape in the body:

| image | HTTP | stored column |
|---|---|---|
| v2.260720.1 (current release) | 500 (invalid byte sequence 0x00) | NULL (write failed) |
| this fix | 200 | "AAABBB" — NUL stripped, no null byte in column |

So media content now reaches the agent instead of the placeholder.

Found while wiring an Agno AgentOS agent to Omni: a PDF whose extracted text contained a NUL byte produced `[media processing unavailable]` despite successful extraction.

### DIFF
--- a/packages/api/src/services/messages.ts
+++ b/packages/api/src/services/messages.ts
@@ -24,6 +24,21 @@ import {
 } from '@omni/db';
 import { and, count, desc, eq, gte, ilike, inArray, lte, or, sql } from 'drizzle-orm';
 
+import { sanitizeText } from '../utils/utf8';
+
+// Pre-processed media text (extraction / vision / transcription) can carry NUL
+// bytes; Postgres text columns reject 0x00, so strip them before any write.
+function sanitizeMediaText(data: Partial<NewMessage>): Partial<NewMessage> {
+  const out: Partial<NewMessage> = { ...data };
+  if (typeof out.transcription === 'string') out.transcription = sanitizeText(out.transcription) ?? '';
+  if (typeof out.imageDescription === 'string') out.imageDescription = sanitizeText(out.imageDescription) ?? '';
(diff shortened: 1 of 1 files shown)


vasconceloscezar:
Thanks @gemini-code-assist — addressed all points in the amended commit:

1. **`?? ''` instead of `?? <original>`** in all four `update*` methods (cleaner, and clears the column even if the input is all-NUL).
2. **`create()`** now sanitizes the pre-processed media fields on insert/sync (`sanitizeText(options.transcription)` etc. — kept undefined-preserving so absent fields stay NULL rather than becoming `''`).
3. **generic `update()`** now runs `data` through a small `sanitizeMediaText()` helper before `.set(...)`, so partial updates carrying pre-processed content are covered too.

`turbo typecheck --filter=@omni/api` passes and biome is clean.

PR #885, opened 2026-07-31
feat(whatsapp-flows): API-first flow CRUD + encrypted data_exchange endpoint

vasconceloscezar:
## Summary

Ships the full WhatsApp Flows lifecycle over the API, including the dynamic (endpoint-backed) mode that was explicitly out of scope until now (`senders/flow.ts` used to say so).

Born from a live debugging session on `khal-ai-demo`: we drove navigate flows end-to-end via the Graph API, hit every silent Flow JSON pitfall (`data_api_version` without endpoint = "an error occurred" with no trace; `RichText` placement/type rules), and turned the findings into API surface, local validation and tests.

### A. Navigate CRUD DX
- New routes: `GET/PUT/DELETE /instances/:id/whatsapp-flows/:flowId`, `POST .../:flowId/deprecate`
- `create`/`update` **always surface Meta's `validation_errors`** (previously swallowed)
- **Local Zod validation** (`FlowJsonSchema` + `validateFlowJson` in `@omni/core`) rejects the known killers before any Graph API round-trip (422 with precise issues)
- Client: `updateFlowAssets` (multipart), `updateFlowMetadata` (`endpoint_uri` is a flow *property*, not Flow JSON), `getFlow`, `deprecateFlow`

### B. data_exchange (send/receive mid-flow)
- Public endpoint `POST /api/v2/channels/whatsapp-cloud/flows/data/:instanceId` with Meta's full contract: HMAC check (**432**), RSA-OAEP + AES-128-GCM decrypt (**421**), bad token (**427**), flipped-IV encrypted responses, plain+encrypted ping, error-notification ack
- **Keys**: `whatsapp_flow_keys` table (migration 0046; private key sealed via `sealCredentialField` under the instance's tenant), `POST/GET .../whatsapp-flows/keys` to generate/rotate/register with Meta and surface `signature_status` (MISMATCH = the 421-loop explainer)
- **Resolver**: in-process `FlowResolverRegistry` (synchronous by design — NATS has no request/reply), 8s budget, degrades to an in-screen error, never hangs; `flow.data_exchange` core event for observability
- **Structured flow tokens** `omni.<flowId>.<uuid>` from `sendFlow` — the decrypted payload carries no flow id, the token is how the endpoint knows which flow it serves

### C. DX + docs
- `@omni/sdk`: hand-authored fluent Flow JSON builder (`flow().screen(...).build()`) enforcing the same rules at build time
- OpenAPI registration for all flows routes; `types.generated.ts` regenerated
- `docs/channels/whatsapp-flows.md` (lifecycle, gotchas, key setup, status-code table)

## Verification
- 6200+ tests green (pre-push full suite): crypto round-trip vs a Meta-shaped client, handler status codes (200/404/421/427/432 + ping + error-notification), validator accept/reject on the live-tested flow, resolver token fallback, route/scope/ownership/RLS gates
- `make typecheck` (23/23) + biome clean
- Migration applied on a disposable database (47/47) and dropped
- Live drive of the dynamic path (keys → dynamic flow → INIT/data_exchange via ngrok) pending an API restart — follow-up on the session

## Notes for review
- Public route declared `public-by-contract` in route-ownership (double-authenticated: HMAC + only the registered key decrypts)
- `whatsapp_flow_keys` has **no** `tenant_id` column on purpose — tenancy derives via `instance_id` (the `whatsapp_templates` precedent); sealing still uses the instance's tenant
- Migration 0046 hand-written per the 0044/0045 precedent (snapshot drift keeps drizzle-kit generate interactive)

### DIFF
--- a/docs/channels/whatsapp-flows.md
+++ b/docs/channels/whatsapp-flows.md
@@ -1,0 +1,158 @@
+# WhatsApp Flows
+
+WhatsApp Flows are in-chat interactive forms/screens (Meta Cloud API channel
+only). Omni supports the full lifecycle: create/update/publish flows via API,
+send them as interactive messages, receive completions, and — for **dynamic
+flows** — host Meta's encrypted data-exchange endpoint so screens can be
+resolved server-side mid-flow.
+
+## Two flavors
+
+| | Static (`navigate`) | Dynamic (`data_exchange`) |
+|---|---|---|
+| Screens | All in Flow JSON, resolved on-device | Decided per-step by omni's data endpoint |
+| Server calls during the flow | None | `INIT` → screen, each submit → next screen |
+| Requirements | Published/draft flow | + encryption keys + `META_FLOWS_PUBLIC_BASE_URL` + a registered resolver |
+| Result delivery | `nfm_reply` webhook at completion | Same, plus per-step `flow.data_exchange` events |
+
+## Management API
+
+All routes live under `/api/v2/instances/:id/whatsapp-flows` (scope:
+`instances:read`/`instances:write`).
+
+```
+GET    /                       list flows
+POST   /                       create  { name, categories, flowJson?, publish?, dynamic? }
+GET    /:flowId                status + validationErrors + endpointUri + preview
+PUT    /:flowId                update  { flowJson?, name?, categories?, dynamic? }
+DELETE /:flowId                delete (DRAFT only)
+POST   /:flowId/publish        publish
+POST   /:flowId/deprecate      retire a published flow
+GET    /:flowId/preview        browser preview URL
+POST   /send                   send  { to, flowId|flowName, cta, bodyText, ..., flowAction? }
+POST   /keys                   generate + register encryption keys (repeat = rotate)
+GET    /keys                   key status (local presence + Meta signature_status)
+```
+
+Create/update run **local Flow JSON validation** (`validateFlowJson` in
+`@omni/core`) before contacting Meta — catching the errors Meta reports only
+after an upload round-trip (or not at all) — and always return Meta's
+`validationErrors` in the response.
+
+### Flow JSON gotchas (enforced locally)
+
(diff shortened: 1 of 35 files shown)


vasconceloscezar:
**Live validation complete** ✅ Drove the dynamic (data_exchange) path end-to-end on the real WABA (`khal-ai-demo` via ngrok):

```
08:53:24  INIT           → endpoint serves START           (200, ~27ms)
08:53:32  data_exchange  → CONFIRM w/ server-built summary (200)
08:53:34  data_exchange  → SUCCESS                         (200)
08:53:36  nfm_reply      → {nome, sabor, confirmado, flow_token} → message.received → agent dispatched
```

Structured token `omni.<flowId>.<uuid>` round-tripped intact. Keys were generated/registered via `POST .../whatsapp-flows/keys`, the flow was created/published/sent entirely through the new API routes.

Field notes baked into the last commit: non-string `extension_message_response.params` values are silently dropped by the client; `flow.>`/`template.>`/`channel.>` had no JetStream subject coverage (pre-existing latent 503s for template/channel events); debugging detour was an unrelated infra issue (the Meta app's webhook callback URL had been repointed to a teammate's tunnel — diagnose with `GET /{app-id}/subscriptions`).

vasconceloscezar:
**GitGuardian finding triaged — false positive, no real secret.** The flagged string was the Meta flow id of the demo flow (`1381873504009404`) inside a test fixture. Flow ids are public resource identifiers (they ride in every flow message payload) and are unusable without the WABA access token.

Audited both PR commits for actual secrets — zero hits for the app secret, `omni_sk_*` keys, Meta access tokens (`EAA…`), verify token or tunnel URLs; `.env` is gitignored and was never committed.

Commit 3f06ebe6 swaps the id for an obviously synthetic one so the detector stays quiet. The incident (id 35622435) can be resolved as false positive in the GitGuardian dashboard.

PR #905, opened 2026-08-13
feat(hooks): retire the SessionStart 300s health hook (hooks-v2#retire)

namastex888:
## hooks-v2#retire — omni leg

Retires the plugin's SessionStart health hook (`omni-runner.js health`, `timeout: 300`) that auto-installed omni and auto-recovered the bridge on every session start.

The health probe moves to `genie doctor` (companion genie PR, automagik-dev/genie — the two land together). The hooks file stays as an empty `hooks: {}` so plugin packaging keeps its shape.

**OPEN-ONLY — never merge from the worker side; the installed-marketplace re-sync is a separate morning follow-up after this merges.**

## Summary by CodeRabbit

* **Changes**
  * Removed the Omni plugin’s automatic health check at session startup.
  * Health diagnostics are now handled through the `genie doctor` command.

### DIFF
--- a/.genie/wishes/omni-plugin-revamp/WISH.md
+++ b/.genie/wishes/omni-plugin-revamp/WISH.md
@@ -33,7 +33,9 @@ encodes that shape in CI gates so it cannot silently rot again.
   `omni-feature-implementor`
 - Fold `plugins/omni/rules/omni-agent.md` into the omni-agent skill; delete it
 - Simplify the SessionStart hook to a probe (no auto-install); shrink
-  `omni-runner.js` to ≤ 80 lines
+  `omni-runner.js` to ≤ 80 lines — SUPERSEDED (2026-08-13): hooks-v2#retire
+  removed the SessionStart hook entirely; the health probe moved to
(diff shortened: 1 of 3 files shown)


vasconceloscezar:
Post-merge follow-up from an independent review (2 low-severity nits that survived the merge):

**1. `WISH.md` is now internally self-contradictory.** The retirement added `SUPERSEDED` annotations in a couple of spots but left the old acceptance criteria intact, so the wish now contradicts the shipped change:
- `omni-runner.js` is asserted to be `≤ 80 lines` (it's ~432) in the plan + verification gate.
- A criterion still says `SessionStart still reports CLI presence/absence on a fresh session` — directly contradicting the hook's removal.

Worth striking through / annotating those superseded criteria so the doc doesn't assert behavior that no longer exists.

**2. Cross-repo sequencing.** The health probe replacement lives in the companion `genie doctor` PR (genie #2780). Since this merged, please confirm #2780 has landed — until it does, fresh installs have no health/auto-recovery probe on either surface (the marketplace re-sync keeps existing installs on the old hook in the interim).

Neither blocks anything now; just tracking so they don't get lost.

PR #931, opened 2026-08-30
fix(api): whatsapp-business persisted config on generic connect/restart and direct sends

### LINKED ISSUE #894: whatsapp-business: instance-persisted config not reaching the plugin outside the dispatcher path (restart bricks the instance; messageFormatMode ignored on direct sends)
Two related gaps where **instance-persisted config never reaches the whatsapp-business plugin** on non-dispatcher API paths. Both observed live on a v2.260804.x deployment (2026-08-05), while operating a whatsapp-business instance via the v2 API.

## Bug 1 — `POST /instances/:id/connect` (and `/restart`) does not hydrate persisted Meta credentials → restart bricks a connected instance

**Repro:**
1. Have a connected whatsapp-business instance (Meta config persisted: `metaAccessToken`, `metaPhoneNumberId`, `metaWabaId` all set on the row).
2. `POST /api/v2/instances/:id/restart` (or `/connect` with empty body).

**Actual:** `500 CONNECTION_FAILED — "metaAccessToken is required to connect a whatsapp-business instance"`, and the running instance is left **disconnected** (subsequent sends fail with `502 CHANNEL_SEND_FAILED — WhatsApp Cloud instance not connected`). Passing `{"metaAccessToken": ..., "metaPhoneNumberId": ...}` in the generic connect body doesn't help — the route never forwards it.

**Why:** `packages/api/src/routes/v2/instances.ts` — `buildConnectConnectionOptions()` threads persisted credentials for baileys/telegram/slack/twilio/gupshup, but has no branch for the Meta fields, so `plugin.connect()` receives `config.credentials = {}`. The restart path (`instances.ts:1345`) calls `plugin.connect(id, { credentials: {}, ... })` directly. This is the same class of bug fixed for hermes in `d6721c95` / `0d933a06` ("pass persisted hermes credentials on instance restart") — whatsapp-business was left out.

**Workaround found:** `POST /api/v2/instances/:id/whatsapp-cloud/connect` with the token from the instance row reconnects fine. But any generic restart (including whatever calls `plugin.connect` on process boot, if it uses the same empty-credentials path) leaves the channel down until someone does this manually.

**Expected:** generic connect/restart hydrates `metaAccessToken` / `metaPhoneNumberId` / `metaWabaId` (+ optional appId/businessId/apiVersion) from the instance row, exactly like hermes does post-fix.

## Bug 2 — `POST /messages/send` ignores `instance.messageFormatMode` → direct sends always convert markdown

**Repro:**
1. `PATCH /api/v2/instances/:id` with `{"messageFormatMode": "passthrough"}` (200, persisted — GET confirms).
2. `POST /api/v2/messages/send` with `text: "*bold*"` on that whatsapp-business instance.

**Actual:** device shows _bold_ in italic — the text went through `markdownToWhatsApp` (`*x*` → `_x_`) despite passthrough being configured.

**Why:** the send route (`packages/api/src/routes/v2/messages.ts`, send-text handler) builds `metadata: { mentions, replyContext }` and never threads `instance.messageFormatMode`. The plugin defaults it: `message.metadata?.messageFormatMode ?? 'convert'` (`packages/channel-whatsapp-business/src/plugin.ts:983` on `8176fcbf`). The agent-responder/agent-dispatcher paths DO thread it — so the same instance behaves differently depending on whether the message is an agent reply or a direct API send. Presumably the same applies to the other `/messages/send/*` variants and to interactive (`content.buttons`) bodies.

**Expected:** the send routes resolve `messageFormatMode` from the instance row (they already load the instance for channel/recipient resolution) and pass it in `metadata`, so instance config governs both paths consistently.

---
Happy to split this into two issues if you prefer; filing together because the root theme is the same — per-instance config not reaching the plugin outside the dispatcher path.

vasconceloscezar:
Fixes #894 — both gaps where instance-persisted whatsapp-business config never reached the plugin outside the dispatcher path.

## Bug 1 — restart/connect bricked the instance

`POST /instances/:id/connect` and `/restart` built plugin options without the `meta*` fields, so `plugin.connect()` threw `metaAccessToken is required` **after** the disconnect had already happened, leaving the channel down until a manual `whatsapp-cloud/connect`.

- `buildConnectConnectionOptions` / create route now thread `metaAccessToken` / `metaPhoneNumberId` / `metaWabaId` (+ optional `metaAppId`, `metaBusinessId`, `metaApiVersion`, `metaDisplayPhoneNumber`, `metaConnectionMethod`) from the instance row.
- New `applyWhatsAppBusinessConnectionOptions` helper (narrowed input, same shape as the hermes fix in d6721c95/0d933a06) reused by the restart path.
- Boot auto-reconnect (`instance-monitor.ts`) already did this via `applyWhatsAppBusinessOptions` — the API routes were the only paths left out.

## Bug 2 — `messageFormatMode` ignored on direct sends

`POST /messages/send` never threaded `instance.messageFormatMode` into the outgoing metadata, so plugins fell back to their `'convert'` default and converted markdown even on `passthrough` instances — direct API sends behaved differently from agent replies on the same instance.

- The send route now resolves the mode from the already-loaded instance row and passes it in `metadata` (omitted when unset, so plugins keep their `convert` default). Interactive (buttons/list) bodies share the route, so they're covered.
- Other `/send/*` variants are unaffected: plugins only apply `messageFormatMode` to `content.type === 'text'`.

## Tests

- `instances-whatsapp-business-connect.test.ts` — connect hydrates required + optional Meta fields, omits keys when none persisted, restart reconnects with credentials after disconnect.
- `messages-send-format-mode.test.ts` — passthrough/convert threading, key omitted when unset, buttons bodies covered.

`bun test` (v2 route suite: 309 pass), `tsc --noEmit`, `biome check --error-on-warnings` all green.


vasconceloscezar (commented):
## Code review

**No correctness defects found.** Reviewed both source hunks, all four `plugin.connect()` call sites in `packages/api/src`, the `messageFormatMode` readers across every channel plugin, and ran the two new test files (8 pass).

The core fix checks out:

- Meta creds belong in `options` — the plugin reads `creds.metaX ?? opts.metaX` and the generic connect/restart routes pass `credentials: {}`. The required trio (`metaAccessToken` / `metaPhoneNumberId` / `metaWabaId`) is exactly what `connect()` throws on.
- All `plugin.connect()` call sites are now covered: instance-monitor, create, connect, restart, and the whatsapp-business route.
- Explicitly passing `messageFormatMode: 'convert'` is a no-op for every reader (`?? 'convert'` everywhere; discord compares against `'passthrough'`), so only `passthrough` instances change behavior — as intended. Other `/send/*` routes are unaffected: telegram applies `formatMode` only to text/poll/button content, and baileys/whatsapp-business/hermes/slack gate on `content.type === 'text'`. `sendTextSchema` has no caller-supplied `metadata`, so nothing is being overridden.

Three low-severity observations inline — none blocking.

packages/api/src/routes/v2/instances.ts
```diff
@@ -1346,6 +1395,14 @@ function buildConnectConnectionOptions(
     hermesPassword: body.hermesPassword ?? instance.hermesPassword,
     hermesMediaId: body.hermesMediaId ?? instance.hermesMediaId,
     hermesTemplateNamespace: body.hermesTemplateNamespace ?? instance.hermesTemplateNamespace,
+    metaAccessToken: instance.metaAccessToken,
+    metaPhoneNumberId: instance.metaPhoneNumberId,
+    metaWabaId: instance.metaWabaId,
+    metaAppId: instance.metaAppId,
+    metaBusinessId: instance.metaBusinessId,
+    metaApiVersion: instance.metaApiVersion,
```
vasconceloscezar, line 1370: **`metaApiVersion` sourcing (low).** This threads the persisted snapshot, but the canonical `POST /:id/whatsapp-business/connect` builds its client from `readMetaAppEnv()` (`process.env.META_GRAPH_API_VERSION`), and the column's own doc comment says "Snapshot of Graph API version used at provisioning. Runtime uses META_GRAPH_API_VERSION env."

Scenario: an instance provisioned at `v25.0` keeps `v25.0` in the row; an operator bumps `META_GRAPH_API_VERSION` after Meta sunsets v25; `POST /instances/:id/restart` then re-pins the live client to the dead version while the canonical connect route would have used the new one.

This matches boot auto-reconnect (`packages/api/src/plugins/instance-monitor.ts:270`), so it's a pre-existing pattern being extended rather than a regression. But if the env is meant to win at runtime, both should read `readMetaAppEnv().apiVersion` with the row as fallback.

packages/api/src/routes/v2/instances.ts
```diff
@@ -861,6 +902,14 @@ instancesRoutes.post('/', zValidator('json', createInstanceSchema), async (c) =>
     hermesPassword: instance.hermesPassword,
     hermesMediaId: instance.hermesMediaId,
     hermesTemplateNamespace: instance.hermesTemplateNamespace,
+    metaAccessToken: instance.metaAccessToken,
```
vasconceloscezar: **Unreachable hydration (low).** Unlike `hermes*` (line 187) and `twilio*` (line 176), `createInstanceSchema` has no `meta*` fields — so `instance.meta*` is always `null` immediately after `services.instances.create(data)` and these eight assignments are dead here.

Harmless, but it reads as if a whatsapp-business instance can be created already-credentialed when it cannot; credentials only arrive via the whatsapp-cloud connect/OAuth route.

packages/api/src/routes/v2/instances.ts
```diff
@@ -603,6 +611,36 @@ function applyHermesConnectionOptions(
...
 }
 
+/**
+ * whatsapp-business (Meta Cloud) persisted credentials — the plugin's
+ * `connect()` reads these from `config.options` (same keys as
+ * `config.credentials` in the whatsapp-cloud connect route). Input type is
+ * narrowed so the restart path can pass the instance row directly (GH #894 —
+ * same shape as the hermes fix in d6721c95/0d933a06).
+ */
+function applyWhatsAppBusinessConnectionOptions(
```
vasconceloscezar: **Duplication (low).** This is byte-identical to `applyWhatsAppBusinessOptions` in `packages/api/src/plugins/instance-monitor.ts:258` — the same eight truthiness-gated assignments. The two will silently drift when Meta adds a field; one of them should be the shared implementation.

Noting that the same duplication already exists for hermes, so this is consistent with repo precedent rather than a new smell.

packages/api/src/routes/v2/instances.ts
vasconceloscezar: Agreed — implemented in 1432f7f5. The hydration helper now resolves the version env-first: `process.env.META_GRAPH_API_VERSION ?? input.metaApiVersion` (read at call time, same as `readMetaAppEnv()`), so an operator bump wins over the stale provisioning snapshot. Because the helper is now shared with `instance-monitor.ts` (see the duplication thread), boot auto-reconnect gets the same env-first behavior — covering both paths you pointed at. Added a regression test (`META_GRAPH_API_VERSION env beats the persisted provisioning snapshot`) that sets the env var and asserts the connect options carry it.

Note `buildClientFromInstance` in `routes/v2/whatsapp-business.ts:105` still builds row-first clients for the template/flows routes — left untouched as it predates this PR and is a separate consistency question.

packages/api/src/routes/v2/instances.ts
vasconceloscezar: Valid — removed in 1432f7f5. You're right that `createInstanceSchema` has no `meta*` fields, so the row can never be credentialed at create and those eight assignments were dead. The create call site now carries a short comment stating exactly that (credentials only arrive via the whatsapp-cloud connect/OAuth route), so the next reader doesn't re-add them "for parity" like I did.

packages/api/src/routes/v2/instances.ts
vasconceloscezar: Implemented in 1432f7f5 — extracted to `packages/api/src/lib/whatsapp-business-connection.ts` (`applyWhatsAppBusinessConnectionOptions` + `WhatsAppBusinessConnectionSource`), and both `routes/v2/instances.ts` (generic connect + restart) and `plugins/instance-monitor.ts` (boot reconnect, replacing its local `applyWhatsAppBusinessOptions`) now import it. The shared implementation is also where the env-first `metaApiVersion` fix lives, so the two call sites can't drift on that either.

Left the hermes/twilio/gupshup duplication as-is — same smell, but deduplicating them here would be scope creep for this PR.

PR #932, opened 2026-08-30
fix(dispatcher): surface messages queued behind an active run (#920)

### LINKED ISSUE #920: Dispatcher serializa runs por chat sem nenhum sinal de fila — indistinguível de mensagem perdida
## Resumo

O dispatcher **serializa os runs por chat** e não emite sinal nenhum enquanto uma mensagem espera. Com provider de tarefa longa (`claude-code`, ~16 min por run no nosso caso), quem mandou vê **silêncio total** — sem log, sem reação, sem mensagem no canal — e conclui que quebrou.

Não é perda de mensagem: a fila funciona e entrega. É **falta de observabilidade da fila**, e o efeito prático é indistinguível de um bug.

## Repro

Instância Slack + provider `claude-code`. Mandar duas menções no mesmo canal com ~90s de intervalo (threads diferentes, o que não importa — o chat é o mesmo).

```
13:43:13  Dispatching to agent   traceId=a94cb99d  activeCount=1 queueDepth=0
13:44:48  message.received       93898d71          ← 2ª menção: NENHUMA linha de dispatcher
                                                     (as que o filtro rejeita logam "Message did
                                                      not pass reply filter" com a razão — essa não)
13:59:12  Agent response         traceId=a94cb99d  durationMs=946901
13:59:12  Dispatching to agent   traceId=93898d71  ← dispara no MESMO milissegundo em que a 1ª fecha
```

A 2ª mensagem ficou **15 minutos** em espera sem produzir um único log. Diagnostiquei errado como "descarte silencioso" antes de ver o dispatch tardio — que é exatamente o ponto: **não há como saber se a mensagem está na fila ou se sumiu.**

## Por que é pior do que parece

1. **A rejeição pelo filtro tem log; a espera na fila não.** A assimetria faz o caminho mais lento parecer o caminho quebrado.
2. **`reactionAck` não dispara ao enfileirar.** Temos `reactionAck: on` com `{"slack":"eyes"}`. A mensagem que esperou 15 min ficou **sem reação nenhuma** — o único sinal visual que existiria só aparece no dispatch. Verificado via `conversations.replies`: `reações=NENHUMA`.
3. **`queueDepth` sempre reporta 0.** Nos dois `agent_dispatch_queue_started` o campo é `0`, mesmo com uma mensagem esperando — então nem a métrica revela a espera.
4. **`maxConcurrency: 8` engana.** O paralelismo é do pool, mas a serialização por chat é o que manda. Quem lê o log assume folga que não existe.

## Sugestões, em ordem de custo

1. **Logar o enfileiramento** — um `agent_dispatch_queued` com `traceId`, posição e o `traceId` do run que está bloqueando. É a correção mínima e resolve o diagnóstico.
2. **Disparar o `reactionAck` no enfileiramento**, não só no dispatch. O 👀 passa a significar "recebi" em vez de "comecei", que é o que o usuário precisa saber.
3. **`queueDepth` refletir a espera por chat.**
4. **(maior) Sessão por thread no Slack.** Hoje o `thread_ts` é ignorado e todas as threads de um canal compartilham a fila. Threads são unidades independentes de conversa — rodar em paralelo seria o comportamento esperado, e evitaria a serialização na maioria dos casos reais.

Relacionado: #914 (migração pra Agent messaging experience), que tem a mesma raiz de falta de sinal para o usuário durante execução longa.

Disposto a mandar PR do item 1 e 2 se a direção fizer sentido.

vasconceloscezar:
## Summary

Fixes the observability gap from #920: the dispatcher serializes runs per chat, and a message waiting behind a long run (15+ min on `claude-code`) produced **zero signal** — no log, no reaction ack, `queueDepth=0` — indistinguishable from a silent drop.

### Root cause (differs slightly from the issue's guess)

The wait does **not** happen in `AgentDispatchLimiter` (which already logs `agent_dispatch_queue_enqueued`). It happens upstream, in `MessageDebouncer.buffer()`: the entire agent run executes inside the debouncer's `onFlush`, so a message arriving mid-run hits the in-flight hold-back branch and accumulates with a bare `return` — no log, no timer. That's why:

- no dispatcher line appears at receive time ("Dispatching to agent" only logs at flush),
- the limiter's `queueDepth` stays `0` (it never sees the held-back message),
- the `reactionAck` never fires (it runs in `processAgentResponse`, i.e. at flush),
- dispatch fires the same millisecond the blocking run finishes (the flush finally-block re-flush).

### Changes (issue items 1 + 2)

1. **`agent_dispatch_queued` log** when a message buffers behind an in-flight flush, with `traceId`, `queuePosition`, and `blockingTraceId` (the run it's waiting on).
2. **Reaction ack at enqueue** via a new `onQueuedBehindActiveRun` hook wired from the dispatcher — 👀 now means "received", not "run started". Safe w.r.t. the reply filter: `shouldProcessMessage` (which owns "Message did not pass reply filter") runs at receive time, before buffering, so filtered messages never reach this ack. The dispatch-time ack takes over when the run starts; a duplicate reaction add is rejected by the platform and swallowed by `startAck`.

Item 3 (`queueDepth`) needs no limiter change — the limiter's number was accurate for *its* queue; the invisible queue was the debouncer's, now covered by the new log. Item 4 (per-thread Slack sessions) is intentionally out of scope.

### Tests

New unit test covering hook firing, queue position, and blocking traceId; confirms no notification for normal (non-in-flight) buffering. `bun test` on debouncer + dispatcher + limiter suites: 101 pass. Typecheck + biome clean.

Closes #920

### DIFF
--- a/packages/api/src/plugins/agent-dispatcher.ts
+++ b/packages/api/src/plugins/agent-dispatcher.ts
@@ -100,6 +100,7 @@ import {
   type DebounceConfig,
   type DispatchMetadata,
   MessageDebouncer,
+  type QueuedBehindActiveRunInfo,
 } from './message-debouncer';
 import { extractPlatformTimestamp } from './message-persistence';
 import { createSessionStorage } from './session-storage';
@@ -3717,6 +3718,36 @@ function buildAckConfig(instance: DispatchInstance): ReactionAckConfig {
   };
 }
 
+/**
+ * #920: Reaction-ack a message the moment it queues behind an active run.
+ * Without this the only ack fires in processAgentResponse — i.e. at dispatch —
+ * so a message waiting out a long agent run (15+ min on claude-code) shows no
+ * signal at all. Cosmetic and fire-and-forget: the dropped AckHandle means the
+ * ack is removed by its hard timeout, and the dispatch-time ack takes over when
+ * the run finally starts (a duplicate add is rejected by the platform and
+ * swallowed by startAck's catch).
+ */
+async function ackQueuedMessage(info: QueuedBehindActiveRunInfo): Promise<void> {
+  try {
+    const instance = info.message.metadata.resolvedInstance as DispatchInstance | undefined;
+    const messageId = info.message.payload.externalId;
+    if (!instance || !messageId) return;
+    const channel = (info.message.metadata.channelType ?? instance.channel) as ChannelType;
+    const plugin = (await getPlugin(channel)) ?? null;
+    startAck(
+      plugin,
+      createPluginAckProvider(plugin),
+      instance.id,
+      info.chatId,
+      messageId,
+      channel,
+      buildAckConfig(instance),
+    );
+  } catch (error) {
+    log.debug('Failed to ack queued message', { chatId: info.chatId, error: String(error) });
+  }
+}
+
 async function dispatchToAgent(
   services: Services,
   instance: DispatchInstance,
@@ -5707,7 +5738,7 @@ export async function setupAgentDispatcher(
   // Create debouncer for message events
   // (registered module-wide so dispatch paths can consult pending buffers
   //  for the stale-reply gate — see isReplySuperseded)
-  const debouncer = new MessageDebouncer(async (_chatKey, messages) => {
+  const onDebouncedFlush = async (_chatKey: string, messages: BufferedMessage[]) => {
     const firstMsg = messages[0];
     if (!firstMsg) return;
 
@@ -5761,6 +5792,11 @@ export async function setupAgentDispatcher(
     }
 
     await processAgentResponse(services, instance, messages, triggerType, db, eventBus);
+  };
+  const debouncer = new MessageDebouncer(onDebouncedFlush, Date.now, {
+    // #920: a message queued behind an active run can wait many minutes with
+    // no visible signal — ack it at enqueue so 👀 means "received", not "started".
+    onQueuedBehindActiveRun: (info) => void ackQueuedMessage(info),
   });
   activeMessageDebouncer = debouncer;
 
--- a/packages/api/src/plugins/agent-dispatcher.ts
+++ b/packages/api/src/plugins/agent-dispatcher.ts
@@ -3719,31 +3719,58 @@ function buildAckConfig(instance: DispatchInstance): ReactionAckConfig {
 }
 
 /**
+ * #920: Enqueue-time reaction acks, keyed by the debouncer chatKey
+ * (`${instanceId}:${chatId}`). At most ONE live entry per chat — i.e. per
+ * blocking run — so a burst of queued messages costs one ack rate-limit token
+ * (the limiter allows 10/min per instance, shared across all chats), not one
+ * per message. The flush path takes the handle: release() when the batch
+ * dispatches (the dispatch-time ack takes over the same reaction), remove()
+ * when a flush-time gate skips the batch. `at` lets a stale entry (flush never
+ * came — e.g. debouncer cleared on shutdown) expire instead of blocking the
+ * chat's queue acks forever.
+ */
+const queuedMessageAcks = new Map<string, { handle?: AckHandle; at: number }>();
+
+/** #920: max ack lifetime — mirrors channel-sdk's DEC-8 hard cap (startAck clamps to it). */
+const QUEUED_ACK_TIMEOUT_MS = 120_000;
+
+function takeQueuedAck(chatKey: string): AckHandle | undefined {
+  const entry = queuedMessageAcks.get(chatKey);
+  queuedMessageAcks.delete(chatKey);
+  return entry?.handle;
+}
+
+/**
  * #920: Reaction-ack a message the moment it queues behind an active run.
  * Without this the only ack fires in processAgentResponse — i.e. at dispatch —
  * so a message waiting out a long agent run (15+ min on claude-code) shows no
- * signal at all. Cosmetic and fire-and-forget: the dropped AckHandle means the
- * ack is removed by its hard timeout, and the dispatch-time ack takes over when
- * the run finally starts (a duplicate add is rejected by the platform and
- * swallowed by startAck's catch).
+ * signal at all. Cosmetic and fire-and-forget. The handle is kept in
+ * queuedMessageAcks so the flush path can hand the reaction over to the
+ * dispatch-time ack (release) or clean it up when the batch is skipped
+ * (remove); see onDebouncedFlush. The ack lifetime uses the SDK's 120s cap
+ * rather than the instance ackTimeoutMs (default 30s): this ack means
+ * "received, waiting" and should survive as long as the SDK allows.
  */
 async function ackQueuedMessage(info: QueuedBehindActiveRunInfo): Promise<void> {
+  const chatKey = `${info.instanceId}:${info.chatId}`;
   try {
+    const existing = queuedMessageAcks.get(chatKey);
+    if (existing && Date.now() - existing.at < QUEUED_ACK_TIMEOUT_MS) return;
     const instance = info.message.metadata.resolvedInstance as DispatchInstance | undefined;
     const messageId = info.message.payload.externalId;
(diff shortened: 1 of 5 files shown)


packages/api/src/plugins/agent-dispatcher.ts
```diff
@@ -3717,6 +3718,36 @@ function buildAckConfig(instance: DispatchInstance): ReactionAckConfig {
...
+ * swallowed by startAck's catch).
+ */
+async function ackQueuedMessage(info: QueuedBehindActiveRunInfo): Promise<void> {
+  try {
+    const instance = info.message.metadata.resolvedInstance as DispatchInstance | undefined;
+    const messageId = info.message.payload.externalId;
+    if (!instance || !messageId) return;
+    const channel = (info.message.metadata.channelType ?? instance.channel) as ChannelType;
+    const plugin = (await getPlugin(channel)) ?? null;
+    startAck(
```
vasconceloscezar: **medium — the discarded `AckHandle`'s hard timeout can strip the *live* dispatch-time ack.**

`startAck` arms a `setTimeout(timeoutMs)` that calls `ackProvider.removeAck(instanceId, chatId, messageId, emoji)`, and `buildAckConfig` defaults `ackTimeoutMs` to 30 000 ms. Because the handle is dropped here, that timer is unstoppable and it acts on the same `(instanceId, chatId, messageId, emoji)` tuple the dispatch-time ack later re-adds.

Scenario: msg2 queues at t=0 (👀 added, orphan timer set for t=30s); the blocking run ends at t=5s; the re-flush dispatches at ~t=8s and `processAgentResponse` adds 👀 to msg2 (it is `messages[0]`); at t=30s the orphan timer fires and unreacts — the 👀 disappears while msg2's own 15-minute run is still executing, which is exactly the "no signal" state #920 is about. The two closures have independent `removed` flags, so neither cancels the other.

Related even without the race: the enqueue ack self-removes after 30s while the message may wait 15 min, so the "received" signal it promises is only 30s long.

vasconceloscezar: **medium — queued-message acks consume the shared per-instance ack rate limit (10/min), starving real dispatch acks instance-wide.**

`ackRateLimiter` in `packages/channel-sdk/src/reaction-ack.ts:63` is a module singleton keyed `ack:${instanceId}` — 10 acks per 60s across *all chats* of an instance. This change adds one token per queued message, and the same message is acked twice (once at enqueue, once at dispatch as `messages[0]`).

Scenario: an impatient user sends 8 messages during a long `claude-code` run — all 8 queue and ack, and the subsequent dispatch acks for that chat *and every other chat on the same instance* silently return the no-op handle for the rest of the minute. The fix can therefore suppress the very ack it exists to guarantee.

Worth gating — e.g. ack only when `queuePosition === 1`, or only once per chat per blocking run.

packages/api/src/plugins/agent-dispatcher.ts
```diff
@@ -3717,6 +3718,36 @@ function buildAckConfig(instance: DispatchInstance): ReactionAckConfig {
...
+/**
+ * #920: Reaction-ack a message the moment it queues behind an active run.
+ * Without this the only ack fires in processAgentResponse — i.e. at dispatch —
+ * so a message waiting out a long agent run (15+ min on claude-code) shows no
+ * signal at all. Cosmetic and fire-and-forget: the dropped AckHandle means the
+ * ack is removed by its hard timeout, and the dispatch-time ack takes over when
+ * the run finally starts (a duplicate add is rejected by the platform and
+ * swallowed by startAck's catch).
+ */
+async function ackQueuedMessage(info: QueuedBehindActiveRunInfo): Promise<void> {
```
vasconceloscezar, line 3754: **low — the enqueue ack fires ahead of the flush-time gates, and nothing removes it on the skip paths.**

The PR description's safety argument covers only `shouldProcessMessage` (reply filter, access, staleness), which runs at receive time. `shouldSkipViaGate` (the LLM respond/skip gate, ~line 5777) and the close-contact / handoff / `agentPaused` gates inside `processAgentResponse` all run at flush.

On the `shouldSkipViaGate` path `processAgentResponse` is never reached, so no ack is added *or removed* there — the queued message keeps a 👀 for the full `ackTimeoutMs` on a message the agent deliberately will not answer. The dispatch path calls `ackHandle.remove()` on all five of its skip branches (lines 3981–4114); this call site keeps no handle and cannot.

Minor and self-healing at 30s, but it makes 👀 mean "received" for messages that are then dropped by design.

vasconceloscezar (commented):
Code review of the diff vs `origin/dev`. The debouncer half is sound: `inFlight` Set→Map preserves `has()` semantics, `queuePosition = buffer.length` is genuinely 1-based (the buffer is deleted synchronously in `flush()` before `inFlight` is set, so nothing stale can be counted), and the re-flush / `clear()` / `flushAll()` paths are unaffected. `bun test` on the debouncer suite passes (26) and `bunx tsc --noEmit` in `packages/api` is clean.

All three findings are on the new enqueue-time ack — two medium, one low. Details inline.

Checked and fine: `blockingTraceId` propagation, `queuePosition` accuracy, `getPlugin`/`ChannelType` handling, the `externalId` null guard (stricter than the dispatch path's `?? ''`), and the new test's double `resolveFirstFlush!()` (the second call lands on the reassigned resolver of the second flush at the 200 ms mark; deterministic).

packages/api/src/plugins/agent-dispatcher.ts
vasconceloscezar: Valid — fixed in bf748765. Added `AckHandle.release()` to channel-sdk: it clears the auto-removal timer and disarms the handle **without** touching the reaction. The enqueue handle is now kept in a module-level map keyed by the debouncer chatKey, and `onDebouncedFlush` takes it and calls `release()` right before `processAgentResponse` — so the orphan timer no longer exists by the time the dispatch-time ack re-adds the tuple, and can never strip it. Covered by a new SDK test (`release() disarms the timer without removing the reaction`).

On the related point: the enqueue ack now uses the SDK's 120s DEC-8 cap instead of the instance `ackTimeoutMs` (default 30s), so the "received" signal lasts as long as the ack system allows. Going past 120s would mean changing the DEC-8 cap in channel-sdk, which I left alone — for a 15-min wait the reaction still self-removes at 2 min if the run hasn't finished. If you want the signal to persist for the whole wait, that's a DEC-8 revisit rather than a dispatcher change.

vasconceloscezar: Valid — implemented the gating you suggested, in bf748765: at most ONE enqueue ack per chat per blocking run. The `queuedMessageAcks` map doubles as the gate — the slot is reserved synchronously before the first `await` so two near-simultaneous enqueues (the NATS handler runs with concurrency 5) can't both pass the check, and a stale entry expires after 120s so a chat whose flush never came (e.g. debouncer cleared on shutdown) isn't blocked forever.

Your 8-message scenario now costs 1 token instead of 8. The dispatch-time re-ack for `messages[0]` still spends its own token, same as before this PR — so the steady-state cost is 2 tokens per blocking run (was 1 pre-PR, would have been N+1 with the original patch).

packages/api/src/plugins/agent-dispatcher.ts
vasconceloscezar: Valid — fixed in bf748765. `onDebouncedFlush` now takes the chat's enqueue-ack handle at the top and settles it on every exit:

- `remove()` on the flush-time skip paths — `shouldSkipViaGate`, the empty-batch guard, and the instance-not-found fallback — so a batch the agent deliberately won't answer loses its 👀 immediately instead of lingering for the full timeout;
- `release()` once the batch reaches `processAgentResponse`, whose own handle (same reaction tuple) already calls `remove()` on all five of its skip branches and in its finally-block, so those paths were covered as soon as ownership transfers;
- an exception mid-flush leaves the handle to its own hard-timeout backstop, which is the pre-existing DEC-8 behavior.

PR #933, opened 2026-08-30
fix(agent-replay): stop replaying turns answered via POST /api/v2/messages/send

### LINKED ISSUE #912: bug(agent-replay): turns answered via POST /api/v2/messages/send are replayed on every reconnect — the send route cannot set senderAgentId
## Summary

`AgentReplayService` decides a message was "missed" by looking for a later reply with
`sender_agent_id IS NOT NULL`. Any turn an agent integration answers through
`POST /api/v2/messages/send` — which is the only way to send interactive list/button
messages — is stored with `sender_agent_id = NULL`, because the route does not accept that
field. Those turns therefore look permanently unanswered, and the inbound message is
re-dispatched to the agent on **every** reconnect for the next 24h.

Observed on a self-hosted deployment (`2.2608xx`, `channel-whatsapp-business`): the API
restarted, and ~16 seconds later the agent re-answered a user message that had been fully
answered 4.5 hours earlier — re-sending a list of time-sensitive options (bookable slots)
that had already expired by then. No user input in between.

## Mechanism

1. An agent integration answers a turn by posting an interactive list through
   `POST /api/v2/messages/send`. This is out of the agent-dispatcher reply path — it is a
   plain API send, and it is the only route that can produce list/button messages.
2. `routes/v2/messages.ts` does not expose `senderAgentId`. The field exists on
   `services/messages.ts` (`SendMessageOptions`), but no HTTP caller can set it, so the
   outbound row lands with `sender_agent_id = NULL`.
3. On restart the instance reconnects and emits `instance.connected`;
   `plugins/event-listeners.ts` calls `AgentReplayService.onInstanceConnect` fire-and-forget.
4. `services/agent-replay.ts::replayMissedMessages` selects inbound messages from the last
   `MAX_REPLAY_WINDOW_MS` (24h) that pass this guard:

   ```sql
   -- Skip messages that already have an agent reply after them in the same chat
   NOT EXISTS (SELECT 1 FROM messages AS reply
               WHERE reply.chat_id = messages.chat_id
                 AND reply.sender_agent_id IS NOT NULL
                 AND reply.platform_timestamp > messages.platform_timestamp
                 AND reply.deleted_at IS NULL)
   ```

   A component-answered turn has no such reply, so it passes.
5. Each selected row is re-published as `message.received` and the dispatcher runs the agent
   again, as if the user had just written.

## Reproduction

1. Wire an agent to an instance with replay enabled.
2. Have the user send a message.
3. Answer that turn by posting an interactive list via `POST /api/v2/messages/send`
   (i.e. the integration suppresses the run's text and sends the component itself).
4. Restart the API (or otherwise force the instance to reconnect) within 24h, with no
   further user input.

**Expected:** nothing is replayed — the turn was answered.
**Actual:** the inbound from step 2 is re-dispatched and the agent answers it a second time,
unprompted, from the user's point of view as a proactive message.

## Blast radius

- Every restart replays the last turn of **every** chat whose latest answer was a component.
  For integrations that use interactive messages as the default answer format, that is most
  chats — so a routine restart produces a burst of unsolicited messages.
- It repeats: the condition is not consumed by the replay. As long as the chat's latest
  answer is a component, the same inbound stays eligible on every reconnect for 24h.
- The replayed answer is regenerated from the agent's session history, so anything
  time-sensitive in it (offers, availability, slots, prices) is re-sent stale. In our case
  users were shown times that had already passed.
- `updateLastSeenAt` does not help: the guard, not the window, is what lets these rows
  through.

## Why the current guard cannot work as written

The guard tries to answer "was this turn answered?" using a field that only the dispatcher's
own reply path sets. Any answer produced out-of-band is invisible to it. Today the send
route cannot even opt in, so an integration has no way to mark its own outbound as agent
authored.

## Suggested fixes (not mutually exclusive)

1. **Let the send route carry authorship.** Accept an optional `senderAgentId` (or
   `sentBy: "agent"`) on `POST /api/v2/messages/send` and persist it. Smallest change, and
   it makes the existing guard correct for integrations that answer out-of-band.
2. **Make the guard provenance-agnostic.** Treat *any* outbound message (`isFromMe = true`)
   after the inbound as evidence that the turn was handled, or track handled-ness explicitly
   (e.g. a `dispatched_at` / `handled_at` marker on the inbound row written when the
   dispatcher completes) instead of inferring it from a later row.
3. **Cap the harm.** A much shorter default replay window, and/or replay only the newest
   unanswered inbound per chat rather than every row in 24h.

Option 2 seems the sturdier one: it removes the coupling between "the turn was handled" and
"who happened to send the reply".

## Related

- #411 — same symptom class (customer-visible side effects re-firing on restart) but a
  different mechanism; that one was about durable subscriber positioning, this one is the
  explicit replay service and survives `startFrom: 'new'`.
- #406 — prior bug from the same field not being propagated (`senderAgentId` dropped by
  channel plugins).

## Notes

Code references were read from a checkout at `2.260731.x`; the deployment where this was
observed runs `2.2608xx`. Line numbers omitted since they drift — the symbols above should
locate it.

vasconceloscezar:
Fixes #912.

## Problem

`AgentReplayService` decided a turn was "missed" by looking for a later reply with `sender_agent_id IS NOT NULL` — a field only the dispatcher's own reply path sets. Turns answered out-of-band through `POST /api/v2/messages/send` (the only route that can produce interactive list/button messages) land with `sender_agent_id = NULL`, so they looked permanently unanswered and the inbound was re-dispatched to the agent on **every** reconnect for 24h, re-sending stale, time-sensitive answers.

## Changes

Both suggested fixes from the issue (they compose):

**1. Provenance-agnostic replay guard** (`services/agent-replay.ts`)
The `NOT EXISTS` guard now keys on `reply.is_from_me = true` instead of `reply.sender_agent_id IS NOT NULL`: any outbound message after the inbound is evidence the turn was handled, regardless of who sent the answer. This alone fixes the replay storm, including for existing rows already in the 24h window at deploy time.

**2. Send routes can carry agent authorship** (`routes/v2/messages.ts`)
`POST /messages/send` and `/messages/send/media` accept an optional `sentBy: 'agent' | 'user'`. `'agent'` resolves the instance's configured agent **server-side** (`sender_agent_id` is an FK to `agents`, so caller-provided ids are not accepted) and threads it into `OutgoingMessage.metadata.senderAgentId`, which channel plugins already echo into the `message.sent` payload for persistence. Beyond replay, this arms the follow-up lifecycle for out-of-band answers (`follow-up-hooks.ts` only arms on agent-authored sends) and fixes attribution in analytics. OpenAPI schemas updated and SDK types regenerated.

## Tests

- `agent-replay.test.ts`: renders the actual messages-query WHERE clause via `PgDialect` and asserts the guard keys on `reply.is_from_me` and no longer references `reply.sender_agent_id`.
- `messages-send-sent-by.test.ts` (new): `sentBy: 'agent'` resolves the instance agent into `metadata.senderAgentId` on both routes; omitted/`'user'`/agent-less instance stay unattributed; unknown values are rejected.

`make typecheck`, `make lint`, and the full `packages/api` suite (2005 pass, 0 fail) are green.

### DIFF
--- a/packages/api/src/routes/v2/messages.ts
+++ b/packages/api/src/routes/v2/messages.ts
@@ -623,6 +623,12 @@ const sendTextSchema = z.object({
     .boolean()
     .optional()
     .describe('Ask the user to share their location (WhatsApp Cloud: native "Send location" button under the text)'),
+  sentBy: z
+    .enum(['agent', 'user'])
+    .optional()
+    .describe(
+      "Authorship of this send. 'agent' attributes the message to the instance's configured agent (persists sender_agent_id), so agent replay and follow-up scheduling treat the turn as agent-answered (#912). Default: unattributed.",
+    ),
 });
 
 // Send media schema
@@ -640,6 +646,12 @@ const sendMediaSchema = z.object({
     .describe('MIME type of the media (e.g. image/gif enables GIF playback for video type)'),
   voiceNote: z.boolean().optional().describe('Send audio as voice note'),
   threadId: z.string().optional().describe('Thread/topic ID (e.g. Telegram forum topic)'),
+  sentBy: z
+    .enum(['agent', 'user'])
+    .optional()
+    .describe(
+      "Authorship of this send. 'agent' attributes the message to the instance's configured agent (persists sender_agent_id), so agent replay and follow-up scheduling treat the turn as agent-answered (#912). Default: unattributed.",
+    ),
 });
 
 function normalizeSendMediaMimeType(data: z.infer<typeof sendMediaSchema>): string {
@@ -648,6 +660,19 @@ function normalizeSendMediaMimeType(data: z.infer<typeof sendMediaSchema>): stri
     return 'audio/ogg; codecs=opus';
   }
   return inferred;
+}
+
+/**
+ * Resolve `sentBy: 'agent'` to the instance's configured agent (#912).
+ * Server-side on purpose: `messages.sender_agent_id` is an FK, so
+ * caller-provided agent ids are not accepted.
+ */
+function resolveSentByAgentId(
+  sentBy: 'agent' | 'user' | undefined,
+  instance: { agentId?: string | null },
+): string | undefined {
+  if (sentBy !== 'agent') return undefined;
+  return instance.agentId ?? undefined;
 }
 
 function buildSendMediaMetadata(data: z.infer<typeof sendMediaSchema>): Record<string, unknown> {
@@ -1187,7 +1212,7 @@ messagesRoutes.post('/send', async (c) => {
     return c.json({ error: { code: 'VALIDATION_ERROR', issues: parsed.error.issues } }, 400);
   }
 
-  const { instanceId, to, replyTo, threadId, mentions, buttons, list, requestLocation } = parsed.data;
+  const { instanceId, to, replyTo, threadId, mentions, buttons, list, requestLocation, sentBy } = parsed.data;
   // Strip internal routing headers and agent directives before sending (GH #300)
   const text = sanitizeOutboundText(parsed.data.text);
   if (!text) {
@@ -1214,6 +1239,8 @@ messagesRoutes.post('/send', async (c) => {
   // Get reply context if replying
   const replyContext = replyTo ? await getReplyContext(services, instanceId, resolvedTo, replyTo) : {};
 
+  const senderAgentId = resolveSentByAgentId(sentBy, instance);
+
   const outgoingMessage: OutgoingMessage = {
     to: resolvedTo,
     threadId,
@@ -1224,7 +1251,7 @@ messagesRoutes.post('/send', async (c) => {
       ...(list ? { list } : {}),
     } as OutgoingContent,
     replyTo,
-    metadata: { ...(mentions ? { mentions } : {}), ...replyContext },
+    metadata: { ...(mentions ? { mentions } : {}), ...replyContext, ...(senderAgentId ? { senderAgentId } : {}) },
   };
 
   // T8: API processed the send request
@@ -1318,6 +1345,8 @@ messagesRoutes.post('/send/media', zValidator('json', sendMediaSchema), async (c
 
   const mediaMimeType = normalizeSendMediaMimeType(data);
 
+  const senderAgentId = resolveSentByAgentId(data.sentBy, instance);
+
   // Build outgoing message
   const outgoingMessage: OutgoingMessage = {
     to: resolvedTo,
@@ -1329,7 +1358,7 @@ messagesRoutes.post('/send/media', zValidator('json', sendMediaSchema), async (c
       filename: data.filename,
       mimeType: mediaMimeType,
     } as OutgoingContent,
-    metadata: buildSendMediaMetadata(data),
+    metadata: { ...buildSendMediaMetadata(data), ...(senderAgentId ? { senderAgentId } : {}) },
   };
 
   // T8: API processed the send request
--- a/packages/api/src/routes/v2/messages.ts
+++ b/packages/api/src/routes/v2/messages.ts
@@ -577,6 +577,15 @@ const updateDeliveryStatusSchema = z.object({
   latestEventId: z.string().uuid().optional(),
 });
 
+// Shared authorship marker for all send routes (#912). 'agent' resolves the
+// instance's configured agent server-side — see resolveSentByAgentId.
+const sentByField = z
+  .enum(['agent', 'user'])
+  .optional()
+  .describe(
+    "Authorship of this send. 'agent' attributes the message to the instance's configured agent (persists sender_agent_id), so agent replay and follow-up scheduling treat the turn as agent-answered (#912). The response echoes the resolved senderAgentId (null when the instance has no configured agent). Default: unattributed.",
+  );
+
 // Mention schema - supports both WhatsApp and Discord formats
 const MentionSchema = z.object({
   id: z.string().min(1).describe('User/role ID to mention'),
@@ -623,12 +632,7 @@ const sendTextSchema = z.object({
     .boolean()
     .optional()
     .describe('Ask the user to share their location (WhatsApp Cloud: native "Send location" button under the text)'),
-  sentBy: z
-    .enum(['agent', 'user'])
-    .optional()
-    .describe(
-      "Authorship of this send. 'agent' attributes the message to the instance's configured agent (persists sender_agent_id), so agent replay and follow-up scheduling treat the turn as agent-answered (#912). Default: unattributed.",
-    ),
+  sentBy: sentByField,
 });
 
 // Send media schema
@@ -646,12 +650,7 @@ const sendMediaSchema = z.object({
     .describe('MIME type of the media (e.g. image/gif enables GIF playback for video type)'),
   voiceNote: z.boolean().optional().describe('Send audio as voice note'),
   threadId: z.string().optional().describe('Thread/topic ID (e.g. Telegram forum topic)'),
-  sentBy: z
-    .enum(['agent', 'user'])
-    .optional()
-    .describe(
-      "Authorship of this send. 'agent' attributes the message to the instance's configured agent (persists sender_agent_id), so agent replay and follow-up scheduling treat the turn as agent-answered (#912). Default: unattributed.",
-    ),
+  sentBy: sentByField,
 });
 
 function normalizeSendMediaMimeType(data: z.infer<typeof sendMediaSchema>): string {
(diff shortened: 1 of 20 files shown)


packages/api/src/services/agent-replay.ts
```diff
@@ -210,12 +210,16 @@ export class AgentReplayService {
...
+              // — the turn was answered, regardless of who sent the answer.
+              // Provenance-agnostic on purpose (#912): answers sent out-of-band
+              // through POST /api/v2/messages/send land with
+              // sender_agent_id = NULL, and a guard keyed on that column
+              // re-dispatched every such turn on every reconnect.
               sql`NOT EXISTS (
               SELECT 1 FROM ${messages} AS reply
               WHERE reply.chat_id = ${messages.chatId}
-                AND reply.sender_agent_id IS NOT NULL
+                AND reply.is_from_me = true
```
vasconceloscezar, line 226: **Medium — the pre-dispatch auto-ack now counts as "the turn was answered", so genuinely unanswered turns are dropped forever.**

`agent-dispatcher.ts:4035` fires `sendTextMessage(channel, instance.id, chatId, agentAckMessage)` *before* `dispatchToAgent`, with the `senderAgentId` parameter omitted (`sendTextMessage`'s 8th arg, `agent-dispatcher.ts:553`). That ack persists as `is_from_me = true, sender_agent_id = NULL` with a `platform_timestamp` after the inbound.

Scenario: an instance with `agentAckMessage` configured (a real column — `schema.ts:462/897`, also overridable per agent-route at `agent-dispatcher.ts:4624`) receives a customer message, the ack goes out, then the API restarts or the WhatsApp socket drops before the agent's real reply lands. On reconnect this new guard is satisfied by the ack alone, so the inbound is skipped and never re-dispatched — permanently. Under the old `sender_agent_id IS NOT NULL` guard it was correctly replayed. This is exactly the crash-mid-turn case replay exists for.

Same holds for `sendErrorFeedback` (`agent-dispatcher.ts:617`, also unattributed): a "sorry, try again" message now marks the turn answered and blocks the retry-on-reconnect. And any unrelated outbound in the chat (an operator note, a broadcast/campaign send) now suppresses replay of inbounds that arrived just before it.

Suggest narrowing the guard to exclude ack/error-class outbounds, or attributing the ack and error sends and keeping a guard that ORs `is_from_me` with an "is a real reply" signal.

packages/api/src/routes/v2/messages.ts
```diff
@@ -650,6 +662,19 @@ function normalizeSendMediaMimeType(data: z.infer<typeof sendMediaSchema>): stri
...
+/**
+ * Resolve `sentBy: 'agent'` to the instance's configured agent (#912).
+ * Server-side on purpose: `messages.sender_agent_id` is an FK, so
+ * caller-provided agent ids are not accepted.
+ */
+function resolveSentByAgentId(
+  sentBy: 'agent' | 'user' | undefined,
+  instance: { agentId?: string | null },
+): string | undefined {
+  if (sentBy !== 'agent') return undefined;
```
vasconceloscezar, line 678: **Low — `sentBy: 'agent'` silently no-ops on instances whose agent comes from an agent route.**

`resolveSentByAgentId` returns `instance.agentId ?? undefined`, and the route returns 201 with no attribution and no warning when it's null. On deployments that route per-chat via `agent_routes` (`agent-dispatcher.ts:4621-4624` shows route config overriding instance config), an integration that sets `sentBy: 'agent'` gets a success response while `sender_agent_id` stays NULL and `follow-up-hooks.ts:97` still bails — i.e. the follow-up-arming half of the fix silently doesn't happen and the caller has no way to detect it.

Either reject with 400 when no agent can be resolved, or log/return a warning.

packages/api/src/routes/v2/messages.ts
```diff
@@ -1329,7 +1358,7 @@ messagesRoutes.post('/send/media', zValidator('json', sendMediaSchema), async (c
       filename: data.filename,
       mimeType: mediaMimeType,
     } as OutgoingContent,
-    metadata: buildSendMediaMetadata(data),
+    metadata: { ...buildSendMediaMetadata(data), ...(senderAgentId ? { senderAgentId } : {}) },
```
vasconceloscezar, line 1386: **Low — `sentBy` coverage stops at `/send` and `/send/media`.**

The other agent-usable out-of-band send routes still build `OutgoingMessage` with no `senderAgentId`: `/send/tts` (2174), `/send/poll` (2771), `/send/location` (1684), `/send/contact` (1601), `/send/sticker` (1510), `/send/embed` (2811), `/send/forward` (2280). An agent that answers a turn with TTS audio or a poll still lands unattributed, so follow-up scheduling never arms and analytics attribution is still wrong for those.

The first finding covers them for replay purposes, so this is incompleteness rather than breakage — but the PR description's "arms the follow-up lifecycle for out-of-band answers" only holds for two of nine send routes.

vasconceloscezar (commented):
Code review of the `sentBy` / replay-guard change.

The plumbing checks out: `metadata.senderAgentId` is echoed into `message.sent` by all eight channel plugins (whatsapp, whatsapp-business, telegram, discord, slack, gupshup, twilio, hermes), `message-persistence.ts` writes it to `sender_agent_id`, `instances.getById` returns the full row so `instance.agentId` is real, the OpenAPI schemas are docs-only (validation uses the local zod schemas, both updated), the SDK regen matches, and `resolveSentByAgentId` mirrors the dispatcher's own `resolveDispatchSenderAgentId` (`agent-dispatcher.ts:3707`) exactly.

Three findings inline — the first is the one that matters.

Also checked and cleared: reaction sends don't emit `message.sent` (`channel-whatsapp/src/plugin.ts:1569` returns without `emitMessageSent`), so `reactionAck` does not trip the new guard; the `blockMessage` send suppressing replay of the blocked inbound is desirable; the correlated subquery can't match its own outer row (outer filters `is_from_me = false`); metadata spread order is key-safe in both routes; and invalid `sentBy` values are rejected 400 on both routes.

packages/api/src/services/agent-replay.ts
vasconceloscezar: Confirmed and fixed in de807367. The guard needed a way to tell a substantive reply from a procedural courtesy send, so the fix threads an explicit marker end-to-end:

- `sendTextMessage` gained a `systemNotice` param; the pre-dispatch auto-ack and `sendErrorFeedback` call sites set it (`agent-dispatcher.ts`).
- All eight channel plugins echo `metadata.systemNotice` into the `message.sent` payload — one line each, parallel to the existing `senderAgentId` echo.
- `buildSentMessageContentFields` merges `omniSystemNotice: true` into the persisted `rawPayload` (no schema migration needed).
- The replay guard adds `AND COALESCE((reply.raw_payload->>'omniSystemNotice')::boolean, false) = false`, so ack/error rows no longer count as answers and the crash-mid-turn case replays again. Rendered-SQL regression test extended; `buildSentMessageContentFields` marker behavior unit-tested.

Two conscious limits: (1) ack/error rows persisted *before* this deploys lack the marker, so for up to 24h after deploy an old ack can still suppress a replay — same exposure the PR already had, converges immediately for new traffic. (2) The 'any unrelated outbound suppresses replay' part (operator note, campaign send) is kept deliberately: an outbound from a human in the chat is evidence the turn is being handled, and re-dispatching the agent into a live human conversation seemed strictly worse. Happy to revisit if you disagree.

packages/api/src/routes/v2/messages.ts
vasconceloscezar: Fixed in de807367, via the log/return-a-warning route rather than a 400: `resolveSentByAgentId` now warn-logs the miss (`sentBy=agent could not be attributed: instance has no configured agent`), and every sentBy-capable route echoes the resolution in the response — `data.senderAgentId` is the attributed agent UUID, or `null` when the instance has no configured agent. The field is present only when the request set `sentBy: 'agent'`, so callers can now detect the silent no-op deterministically. Mirrored in the OpenAPI `MessageResponse` schema and the regenerated SDK types.

I chose not to reject with 400 because on agent_routes deployments the send itself must still go out — failing the customer-visible message over a missing attribution seemed like the worse trade. Resolving through `agent_routes` server-side would be the fuller fix but needs per-chat route resolution in the send path; left out of this PR's scope.

packages/api/src/routes/v2/messages.ts
vasconceloscezar: Fixed in de807367: `sentBy` now covers `/send/sticker`, `/send/contact`, `/send/location`, `/send/tts`, `/send/forward`, `/send/poll` and `/send/embed` in addition to `/send` and `/send/media` — a shared `sentByField` schema fragment, the same server-side `resolveSentByAgentId`, metadata threading into `OutgoingMessage`, and the `senderAgentId` response echo on each. OpenAPI schemas mirrored for the routes that have registered schemas (sticker/contact/location/tts — poll/embed/forward have no OpenAPI registration today); SDK regenerated. Route tests extended with a table-driven pass over sticker/contact/location/poll.

Deliberately left out: `/send/handoff` and `/send/close-contact` (agent-flow primitives with their own chat-state side effects — pause/close — where attribution semantics deserve their own decision), and `/send/reaction` / `/send/presence` (no `message.sent` persistence, nothing to attribute).

Issue #925, opened 2026-08-29
RFC: Omni as event backbone — event contracts, ingress idempotency, causality, and declarative subscription


vasconceloscezar:
# RFC: Omni as event backbone — event contracts, ingress idempotency, causality, and declarative subscription

## Motivation

Today, omnichannel means "text channels" (WhatsApp, Slack, Telegram, Discord). But the architecture already underneath — NATS JetStream, event envelope, automations, webhook sources, DLQ, replay — is a **generic event bus** that only happens to serve messages.

This RFC's thesis: **embrace omni as the platform's event controller**, where any source (GitHub, ClickUp, Drive, calendar, LLM provider billing, meeting minutes, third-party apps via webhook) emits events that agents "listen" to and act on — with deterministic trace-back of the causal chain and state that can be rebuilt/diffed from the journal.

This is not a pivot. It formalizes what the code already half-does, closing the gaps that today prevent plugging in external sources safely.

## Current state (audited in the code)

What already exists and where:

| Piece | Where | State |
|---|---|---|
| Bus | NATS JetStream, subjects `{eventType}.{channelType}.{instanceId}` (`packages/core/src/events/nats/`) | ✅ live, `consumer_offsets` with gap detection |
| Envelope | `OmniEvent` in `packages/core/src/events/types.ts` — `id`, `type`, `payload`, `metadata` (`correlationId`, `traceId`, `timings`, `ingestMode`, `streamSequence`, `envelopeVersion`/`tenantId`) | ✅ versioned, tenant-aware (G5/ADR-0008) |
| Journal | event persistence (`packages/api/src/plugins/event-persistence.ts`) + `payload_storage_config` with per-type retention | ✅ ~349k events today |
| Handler dedup | `processed_events` (PK `event_id + handler`) via `packages/api/src/lib/idempotency.ts` | ✅ but CONSUMER-side only |
| DLQ | `dead_letter_events` with auto/manual retry, `next_auto_retry_at` | ✅ |
| External ingress | `webhook_sources` — name, `expected_headers`, enabled, stats | ⚠️ validates header presence only, no schema, no dedup |
| Routing | `automations` — trigger by type + conditions (11 operators, and/or) + actions (`webhook`, `send_message`, `emit_event`, `log`, `call_agent`) + debounce + priority + `automation_logs` | ✅ but imperative wiring |
| Replay | replay sessions (`packages/core/src/events/replay.ts`), journey tracing via checkpoints | ✅ |
[...]

PR #934, opened 2026-08-30
fix(automations): dispatch call_agent to local claude-code providers

### LINKED ISSUE #929: bug(automations): call_agent cannot dispatch to local claude-code providers (unconditional API-key check + HTTP-only client factory)
## Summary

The `call_agent` automation action cannot dispatch to a `claude-code` provider (`baseUrl: local://claude-code`), even though the same provider serves normal message routing fine. It fails with `Provider <id> has no API key configured`.

Found while dogfooding event-driven automations (context: #925): a custom event → automation → `call_agent` targeting an agent bound to a local claude-code provider.

## Repro

1. Have an agent bound to a claude-code provider (e.g. one that already answers chats via message routing).
2. Create an automation:
   ```
   omni automations create --name t --trigger "custom.x" \
     --action call_agent --action-config '{"agentId":"<agent-uuid>","promptOverride":"pong","responseAs":"r"}'
   ```
3. Trigger with a chat-complete payload (`instanceId`, `chatId`, `senderId` present).
4. Automation log: `call_agent` fails with `Provider 387c34ba-… has no API key configured`.

## Root cause (traced)

`executeCallAgentAction` → `AgentRunner.getClient()` (`packages/api/src/services/agent-runner.ts:342`), which:

1. requires a non-empty `apiKey` unconditionally (`agent-runner.ts:377-386`) — but local claude-code providers legitimately have none;
2. then builds an HTTP client via `createProviderClient({schema, baseUrl, apiKey…})` — which cannot speak `local://claude-code` anyway (the same reason provider health checks on these providers report `protocol must be http:, https: or s3:`).

The message-routing path dispatches claude-code providers through `claude-code-provider.ts` and works; the automation path never reaches it.

## Expected

`call_agent` should route through the same provider dispatch used by message routing, so any provider that can answer a chat can also be invoked by an automation. At minimum, local-schema providers should bypass the API-key requirement and the HTTP client factory.

## Secondary observations (same path)

- `extractAgentCallContext` requires `chatId`/`senderId` from the payload, so a chatless (non-message) event can never invoke an agent even when the agent needs no chat context — worth decoupling as part of #925 G4.
- Provider health checks report `protocol must be http:, https: or s3:` for every `local://` provider (`lastHealthError` on all claude-code providers), so health status is permanently `error` for a class of provider that works.

vasconceloscezar:
## Summary

Fixes #929 — the `call_agent` automation action could not dispatch to `local://claude-code` providers, failing with `Provider <id> has no API key configured` even though the same provider serves normal message routing fine.

## Changes (all in `AgentRunnerService`, the path `call_agent` lands on)

- **`getClient`**: claude-code providers are now built from `schemaConfig` (`projectPath`) via the factory's existing `claude-code` branch, with the API key optional (the SDK falls back to the host's `ANTHROPIC_API_KEY`). The sealed-credential posture is unchanged and now schema-uniform: a sealed envelope that cannot be opened in the current tenant context fails closed for **every** schema; the "no API key configured" 401 still applies to HTTP schemas.
- **`run()`/`stream()` session continuity**: Claude Code resumes by its own session UUID, not the strategy-computed key. Both methods now map key → UUID through the same per-provider session store the message-routing dispatcher uses (`createSessionStorage`, same `provider:<id>:session:<key>` keys, same `computeSessionId` inputs) — so an automation `call_agent` resumes the very session the chat conversation runs in, and persists the returned UUID for the next turn from either path. HTTP MCP servers also get the same `chat_id` URL param the dispatcher appends.

## Tests

- New `agent-runner-claude-code.test.ts`: keyless claude-code client construction (the #929 regression), plaintext key forwarding, sealed-credential fail-closed, fresh-session start + UUID persistence, stored-session resume, and stream-path session capture.
- Existing agent-runner suites (sealed provider, tenant-keyed client cache, runOrStream) pass unchanged: 484 pass / 0 fail across services + dispatcher + automation-actions suites; `tsc` and biome clean.

## Not addressed (secondary observations in #929)

- `extractAgentCallContext` still requires `chatId`/`senderId` (chatless events — deferred to #925 G4).
- Provider health checks still probe `local://` base URLs over HTTP and permanently report `error` for claude-code providers.

### DIFF
--- a/packages/api/src/services/agent-runner.ts
+++ b/packages/api/src/services/agent-runner.ts
@@ -12,6 +12,8 @@ import {
   type IAgentClient,
   ProviderError,
   type ProviderFile,
+  type ProviderSchema,
+  type SessionStorage,
   type StreamChunk,
   createProviderClient,
   isProviderSchemaSupported,
@@ -21,6 +23,7 @@ import type { AgentReplyFilter, AgentSessionStrategy, ChannelType, Instance } fr
 import type { Database } from '@omni/db';
 import { agentProviders, instances, persons } from '@omni/db';
 import { eq } from 'drizzle-orm';
+import { createSessionStorage } from '../plugins/session-storage';
 import { isSealedCredentialField, openCredentialField } from '../tenancy/sealed-credentials';
 import { currentTenantScope, scopedHandle } from '../tenancy/tenant-scope';
 
@@ -315,6 +318,16 @@ async function* processStreamChunks(
 // Agent Runner Service
 // ============================================================================
 
+/**
+ * A provider client plus the schema it was built for. The schema drives
+ * dispatch behavior in run()/stream(): local schemas (claude-code) need
+ * provider-session mapping instead of passing the strategy key through.
+ */
+interface ResolvedProviderClient {
+  client: IAgentClient;
+  schema: ProviderSchema;
+}
+
 export class AgentRunnerService {
   /**
    * Provider clients, keyed `providerId::tenant`.
@@ -327,7 +340,7 @@ export class AgentRunnerService {
    * and its key — would be served to every later tenant. The scope-less legacy
    * path keys on `-`, so it still shares exactly one client per provider.
    */
-  private clientCache: Map<string, IAgentClient> = new Map();
+  private clientCache: Map<string, ResolvedProviderClient> = new Map();
 
   constructor(private db: Database) {}
 
@@ -337,9 +350,9 @@ export class AgentRunnerService {
   }
 
   /**
-   * Get or create an Agno client for a provider
+   * Get or create a provider client for a provider
    */
-  private async getClient(providerId: string): Promise<IAgentClient> {
+  private async getClient(providerId: string): Promise<ResolvedProviderClient> {
     // Check cache
     const cacheKey = this.clientCacheKey(providerId);
     const cached = this.clientCache.get(cacheKey);
@@ -374,27 +387,73 @@ export class AgentRunnerService {
     // fails CLOSED — a null never becomes a bearer token.
     const apiKey = openCredentialField(currentTenantScope()?.tenantId ?? null, provider.apiKey);
 
-    if (!apiKey) {
+    // A sealed credential that cannot be opened in this tenant context is an
+    // error for EVERY schema — the alternative is silently running under the
+    // wrong identity (or the host's ambient key).
+    if (!apiKey && isSealedCredentialField(provider.apiKey)) {
       throw new ProviderError(
-        isSealedCredentialField(provider.apiKey)
-          ? `Provider ${providerId} credential is not available in this tenant context`
-          : `Provider ${providerId} has no API key configured`,
+        `Provider ${providerId} credential is not available in this tenant context`,
         'AUTHENTICATION_FAILED',
         401,
       );
     }
 
+    // claude-code providers run in-process (`local://claude-code`): there is no
+    // HTTP endpoint and legitimately no API key (the SDK falls back to the
+    // host's ANTHROPIC_API_KEY). Build the client from schemaConfig instead of
+    // requiring a bearer credential — this is how `call_agent` automations reach
+    // the same providers that already answer chats via message routing (#929).
+    if (provider.schema === 'claude-code') {
+      const resolved: ResolvedProviderClient = {
+        client: createProviderClient({
+          schema: provider.schema,
+          baseUrl: provider.baseUrl,
+          apiKey: apiKey ?? '',
+          defaultTimeoutMs: (provider.defaultTimeout ?? 600) * 1000,
+          schemaConfig: (provider.schemaConfig ?? {}) as Record<string, unknown>,
+        }),
+        schema: provider.schema,
+      };
+      this.clientCache.set(cacheKey, resolved);
+      return resolved;
+    }
+
+    if (!apiKey) {
+      throw new ProviderError(`Provider ${providerId} has no API key configured`, 'AUTHENTICATION_FAILED', 401);
+    }
+
     // Create client
-    const client = createProviderClient({
+    const resolved: ResolvedProviderClient = {
+      client: createProviderClient({
+        schema: provider.schema,
+        baseUrl: provider.baseUrl,
+        apiKey,
+        defaultTimeoutMs: (provider.defaultTimeout ?? 600) * 1000,
+      }),
       schema: provider.schema,
-      baseUrl: provider.baseUrl,
-      apiKey,
-      defaultTimeoutMs: (provider.defaultTimeout ?? 600) * 1000,
-    });
+    };
 
     // Cache it
-    this.clientCache.set(cacheKey, client);
-    return client;
+    this.clientCache.set(cacheKey, resolved);
+    return resolved;
+  }
+
+  /**
+   * Session store for a claude-code provider, per run.
+   *
+   * Claude Code resumes by its OWN session UUID, not the strategy-computed
+   * key. The mapping key → UUID lives in the same per-provider store the
+   * message-routing dispatcher uses (`ClaudeCodeAgentProvider` +
+   * `createSessionStorage`), and both paths compute the key with
+   * `computeSessionId` over the same inputs — so a `call_agent` automation
+   * resumes the very session the chat conversation runs in.
+   */
+  private claudeSessionStore(instance: RunInstance, providerId: string): SessionStorage {
+    return createSessionStorage(this.db, providerId, undefined, {
+      // Persisted ownership from the loaded instance row — same trusted
+      // derivation the dispatcher supplies (G5, ADR-0008).
+      resolveTenantId: () => instance.tenantId ?? null,
+    });
   }
 
   /**
@@ -477,7 +536,7 @@ export class AgentRunnerService {
       throw new ProviderError('No agent internal ID configured for instance', 'NOT_FOUND', 400);
     }
 
-    const client = await this.getClient(instance.agentProviderId);
+    const { client, schema } = await this.getClient(instance.agentProviderId);
 
     // Format messages with sender name prefix if enabled
     const prefixEnabled = instance.agentPrefixSenderName ?? true;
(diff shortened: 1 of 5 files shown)


packages/api/src/services/agent-runner.ts
```diff
@@ -374,27 +387,73 @@ export class AgentRunnerService {
     // fails CLOSED — a null never becomes a bearer token.
     const apiKey = openCredentialField(currentTenantScope()?.tenantId ?? null, provider.apiKey);
```
vasconceloscezar: **medium — the fix is incomplete for providers that *do* have an API key.**

`openCredentialField` derives the tenant from `currentTenantScope()`, but the `call_agent` path deliberately runs `runOrStream` *outside* every tenant scope — see the header comment in `plugins/automation-actions.ts` and `callAgent` (~line 223), where each read is wrapped in its own `runTenantWorkDb` and the agent run itself is not.

So on this path `currentTenantScope()` is `null`, a sealed `agent_providers.api_key` cannot be opened, and the new fail-closed check just below (line 391) throws `credential is not available in this tenant context`. Issue #929 still reproduces for any claude-code provider with a stored key — the keyless case is fixed, the keyed case is not.

Note the asymmetry this same diff introduces: `claudeSessionStore` (line 455) correctly derives the tenant from `instance.tenantId`, which `getClient` has no access to. The two halves of a single run therefore disagree about which tenant they are in. Consider threading the instance's persisted tenant into the credential open (or resolving the client per-run rather than per-provider) so both halves agree.

packages/api/src/services/agent-runner.ts
```diff
@@ -374,27 +387,73 @@ export class AgentRunnerService {
...
+    return resolved;
+  }
+
+  /**
+   * Session store for a claude-code provider, per run.
+   *
+   * Claude Code resumes by its OWN session UUID, not the strategy-computed
+   * key. The mapping key → UUID lives in the same per-provider store the
+   * message-routing dispatcher uses (`ClaudeCodeAgentProvider` +
+   * `createSessionStorage`), and both paths compute the key with
```
vasconceloscezar: **medium — this doc claim is not accurate, and the divergence is a real behaviour bug.**

> both paths compute the key with `computeSessionId` over the same inputs

The message-routing dispatcher does **not** call `computeSessionId` for claude-code. It goes through `resolveKhalSessionId` (`services/agent-session-identity.ts`, invoked at `plugins/agent-dispatcher.ts:2361` and `:2838`), which differs in two ways:

1. it passes `threadId` into the `per_thread` branch;
2. it returns an explicit `khal_session_id` / `session_id` / `x-khal-session-id` from the raw payload when one is present.

`run()` and `stream()` call `computeSessionId(strategy, senderId, chatId)` with no thread id (lines 550 and 702).

Concrete misbehaviour: on a `per_thread` instance the automation writes and reads `thread:<chatId>:<chatId>`, while the chat path uses `thread:<chatId>:<threadId>`. The automation never resumes the thread's session, and it collapses *every* thread of that chat into a single shared Claude session. The same divergence applies to channels whose payload carries an explicit khal session id.

Routing this path through `resolveKhalSessionId` would make the comment true and fix both cases.

packages/api/src/services/agent-runner.ts
```diff
@@ -650,10 +732,30 @@ export class AgentRunnerService {
...
+    // persisted under the strategy key once the stream completes.
+    let providerSession: string | undefined;
+    const source = (async function* (): AsyncGenerator<StreamChunk> {
+      for await (const chunk of client.stream(request)) {
+        if (chunk.sessionId) providerSession = chunk.sessionId;
+        yield chunk;
+      }
+    })();
+
+    yield* processStreamChunks(source, enableSplit);
```
vasconceloscezar: **low/medium — the stream path can orphan a freshly created Claude session.**

The `upsertSession` below is not wrapped in `try/finally`, unlike the dispatcher's equivalent in `packages/core/src/providers/claude-code-provider.ts:315`, which persists from a `finally` block.

If the consumer breaks out of the `for await` early, or anything throws out of `processStreamChunks` (an abort, or a non-claude-code client that rethrows), the captured `providerSession` is dropped on the floor: the Claude session that was just created is orphaned and the next call starts fresh, silently losing continuity.

`ClaudeCodeClient.stream()` currently swallows its own errors into a final chunk, which limits the exposure today but does not remove it — the early-break case is entirely up to the caller.

packages/api/src/services/agent-runner.ts
```diff
@@ -527,10 +594,19 @@ export class AgentRunnerService {
...
+      // Parity with the dispatcher's claude-code path: HTTP MCP servers get the
+      // conversation identifier appended to their URL.
+      ...(sessionStore && chatId ? { mcpUrlParams: { chat_id: chatId } } : {}),
     };
 
     const response = await client.run(request);
 
+    // Persist the provider session UUID under the strategy key so the next
+    // call — from either dispatch path — resumes this session.
+    if (sessionStore && response.sessionId) {
```
vasconceloscezar, line 661: **low — a best-effort bookkeeping write can discard a successful agent reply.**

This `await sessionStore.upsertSession(...)` runs unguarded *after* a successful agent run. A transient failure — a DB blip, or an RLS `WITH CHECK` refusal on `agent_sessions` when `instance.tenantId` doesn't match the instance's persisted owner — throws out of `run()`. The reply the agent already produced is thrown away and the automation reports failure, for what is only continuity bookkeeping.

Wrapping the persist in `try/catch` + `log.warn` keeps the reply and degrades to "next call starts a fresh session", which is the pre-fix behaviour anyway. The same applies to the stream-path write at line 757.

packages/api/src/services/agent-runner.ts
```diff
@@ -490,6 +549,14 @@ export class AgentRunnerService {
     const sessionStrategy = instance.agentSessionStrategy ?? 'per_chat';
     const sessionId = computeSessionId(sessionStrategy, senderId, chatId);
 
+    // claude-code: resolve the strategy key to the provider's session UUID (or
+    // start a fresh session when none is stored). Other schemas take the key
+    // as-is, unchanged.
+    const sessionStore = schema === 'claude-code' ? this.claudeSessionStore(instance, instance.agentProviderId) : null;
```
vasconceloscezar, line 612: **low — newly reachable cross-path concurrency on one Claude session.**

With this change the automation resumes the *same* Claude Code session UUID that the chat dispatcher uses for that chat. The dispatcher serialises chat dispatches through its own queue, but the automation engine consumer is independent of that queue.

So a `chat.idle_timeout` `call_agent` firing while a user message is being dispatched results in two SDK queries resuming the same session id concurrently, plus two last-write-wins upserts of the mapping row. Previously the automation always ran an unrelated session, so this interleaving is newly reachable — worth a note at minimum, or reusing the dispatcher's per-chat serialisation if session corruption turns out to matter in practice.

vasconceloscezar (commented):
## Code review — `fix(automations): dispatch call_agent to local claude-code providers`

Reviewed `packages/api/src/services/agent-runner.ts` and the new `agent-runner-claude-code.test.ts`. `bunx tsc --noEmit -p packages/api` is clean, the new suite passes 6/6, and neighbouring service suites are unaffected (the only failures in `packages/api/src/services/__tests__` are the two pre-existing MinIO docker-pull tests).

5 findings, none of them a hard blocker, but the first two are worth resolving before merge:

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | medium | `agent-runner.ts:388` | Fix is incomplete for sealed-credential providers — `#929` still reproduces when the provider has an API key |
| 2 | medium | `agent-runner.ts:447` | Session keys do **not** match the dispatcher's for `per_thread` — collapses every thread of a chat into one Claude session |
| 3 | low/medium | `agent-runner.ts:754` | Stream-path `upsertSession` is not in `try/finally`; early break or throw orphans the session |
| 4 | low | `agent-runner.ts:606` | Unguarded session-store write discards an already-produced agent reply on a transient failure |
| 5 | low | `agent-runner.ts:555` | Newly reachable: automation and chat dispatch can resume the same Claude session UUID concurrently |

Details inline.

packages/api/src/services/agent-runner.ts
vasconceloscezar: Fixed in 9bb9d86f. `getClient` now accepts an `openedForTenantId` and `run()`/`stream()` thread `instance.tenantId` into it — the same persisted-ownership derivation `claudeSessionStore` already used, and the same threaded-over-ambient rule as the dispatcher's `openclawPoolKey`. The credential open and the client cache key are now computed from one resolved tenant, so the two halves of a run agree; with no threaded value (direct callers, legacy world) the ambient-scope behaviour is byte-identical to before. Regression test: "a sealed credential OPENS under the instance's persisted tenant with NO scope active" in agent-runner-claude-code.test.ts.

packages/api/src/services/agent-runner.ts
vasconceloscezar: Fixed the per_thread case in 9bb9d86f, and corrected the comment. `AgentCallContext` and `AgentRunContext` gained an optional `threadId`, extracted from the automation payload (`payload.threadId ?? rawPayload.threadId`) and passed through to `computeSessionId` — a per_thread automation now writes/reads `thread:<chatId>:<threadId>`, matching the dispatcher's legacy key. Tests cover extraction (core actions.test.ts) and the resulting session key (agent-runner-claude-code.test.ts).

I deliberately did NOT route this path through `resolveKhalSessionId`, for two reasons: (1) the runner is the shared legacy path for every schema, and `resolveKhalSessionId` activates canonical khal ids for agno whenever an environment resolves — `resolveKhalEnvironment` falls back to `KHAL_SESSION_ENV`/`KHAL_ENV`/`OMNI_ENV` process env — so adopting it here would silently re-key existing agno sessions on the legacy dispatch/automation paths, well beyond this bug fix; (2) the explicit-khal case needs the message rawPayload, which chat.idle_timeout payloads (the primary call_agent trigger) don't carry. The `claudeSessionStore` doc now states that remaining divergence explicitly instead of claiming parity.

packages/api/src/services/agent-runner.ts
vasconceloscezar: Fixed in 9bb9d86f — the persist now runs from a `finally` around the `yield*` (mirroring `ClaudeCodeAgentProvider.triggerStream`), through the shared `persistProviderSession` helper, so an early consumer break or a mid-stream throw no longer orphans the freshly created session. Test: "stream(): a consumer breaking out early still persists the captured session".

packages/api/src/services/agent-runner.ts
vasconceloscezar: Fixed in 9bb9d86f — both the sync and stream persists go through a new `persistProviderSession` helper that catches and `log.warn`s instead of throwing, so a DB blip or RLS refusal degrades to "next call starts a fresh session" (the pre-fix behaviour) rather than discarding the reply. Test: "a failed session-store write never discards the reply the agent produced".

packages/api/src/services/agent-runner.ts
vasconceloscezar: Acknowledged — took the "note at minimum" option in 9bb9d86f: the `claudeSessionStore` doc now spells out that the automation engine consumer bypasses the dispatcher's per-chat queue, so an idle-timeout call_agent can resume the same session UUID while a chat message is mid-dispatch, and that the mapping upserts converge (both write the identical UUID, last-write-wins on timestamps only). I held off on reusing the dispatcher's per-chat serialisation here: it would couple the runner to dispatcher internals for a race whose practical cost (two concurrent SDK resumes) we haven't yet observed to corrupt anything — the note points there if it does.

PR #935, opened 2026-08-30
fix(api): invalidate cached agent providers on instance/agent config changes

### LINKED ISSUE #906: Instance config (enableAutoSplit, agentTimeout, …) is stale until pod restart — providerCache never invalidated on instance update/restart
## Summary

Updating an **instance** config field that gets baked into the agent provider — most visibly `enableAutoSplit` and the `messageSplitDelay*` family, but also `agentType`, `agentTimeout`, `agentPrefixSenderName`, and the resolved `agentId` — has **no effect until the whole omni-api process restarts**. Neither `PATCH /instances/:id` nor `POST /instances/:id/restart` picks up the change. The value persists correctly in the DB, but the running dispatcher keeps using a stale cached provider.

Found on prod (`omni.hv-cx.khal.ai`, single hermes instance): set `enableAutoSplit=false` via `PATCH /instances/:id` → DB shows `false`, but the agent kept splitting replies into multiple bubbles. A `POST /instances/:id/restart` did **not** fix it either.

## Root cause

`resolveProvider()` caches the built `IAgentProvider` in a module-level `providerCache`, keyed by `${provider.id}:${instance.id}`:

- `packages/api/src/plugins/agent-dispatcher.ts:4432-4470` — `resolveProvider()` returns the cached provider if present, otherwise builds and caches it.
- `packages/api/src/plugins/agent-dispatcher.ts:4217-4223` — `createAgnoProvider()` **bakes instance config into the provider at construction**: `enableAutoSplit: instance.enableAutoSplit ?? true`, plus `agentType`, `timeoutMs`, `prefixSenderName`, and the resolved `agentId`.
- `packages/core/src/providers/agno-provider.ts:85-92` — the split decision reads `this.config.enableAutoSplit`, which was frozen at construction time. So a cached provider keeps the old splitting behavior forever.

The cache is only ever invalidated by **provider** mutations, never by **instance** mutations:

- `packages/api/src/plugins/agent-dispatcher.ts:4482` — `invalidateProviderCache(providerId)` clears entries by `${providerId}:` prefix.
- `packages/api/src/routes/v2/providers.ts:153,173` — called on provider PATCH / DELETE only.
- `packages/api/src/routes/v2/instances.ts:828-865` — the instance PATCH handler calls `services.instances.update(id, data)` and **never invalidates** the provider cache.
- `packages/api/src/routes/v2/instances.ts:1357` — `POST /:id/restart` only does `plugin.disconnect(id)` + `plugin.connect(id, …)`. That recycles the **channel connection**; it does not touch `providerCache` (which lives at the dispatcher module level, independent of channel lifecycle).

The only things that clear `providerCache` for an instance today:
1. A **provider** PATCH/DELETE (`invalidateProviderCache`).
2. Full dispatcher shutdown → process/pod restart (`providerCache.clear()`, `agent-dispatcher.ts:6170`).
3. Test-only reset.

## Impact

Any instance-level field consumed at provider construction is silently stale until a pod restart. Confirmed set (agno schema):
- `enableAutoSplit`
- `messageSplitDelay*` — *(passed per-send via `getSplitDelayConfig(instance)`, so actually fresh on the send path, BUT split only happens when the cached provider produced ≥2 parts, so a stale `enableAutoSplit=true` keeps splitting regardless)*
- `agentType`, `agentTimeout`, `agentPrefixSenderName`, resolved `agentId`

Other schemas (`openclaw`, `claude-code`, `ag-ui`, `a2a`, `nats-genie`) build from `instance` too, so they share the staleness.

This is a footgun: the API returns `200`, the DB reflects the change, `GET /instances/:id` shows the new value — everything looks applied, but runtime behavior doesn't change. Operators reasonably conclude the change is live when it isn't.

## Reproduction

1. Instance with an agno provider, `enableAutoSplit=true`, replies that contain `\n\n`.
2. Send a message → reply arrives split into multiple bubbles.
3. `PATCH /instances/:id` with `{"enableAutoSplit": false}` → `200`, `GET` confirms `false`.
4. Send another message → **still split** (stale cached provider).
5. `POST /instances/:id/restart` → reconnects channel, **still split**.
6. Only a pod restart (or a provider PATCH) makes it stop.

## Workaround (no pod restart)

Trigger provider-cache invalidation by touching the provider:

```
PATCH /api/v2/providers/:providerId   { "name": "<same name>" }
```

This calls `invalidateProviderCache(providerId)`, evicting the cached provider for **all** instances under it; the next inbound message rebuilds it from the fresh instance row.

## Suggested fix

Invalidate the affected provider cache entry when an instance is updated (and on instance restart). Options:

1. In the instance PATCH handler (`instances.ts:828-865`), after `services.instances.update`, if any provider-affecting field changed, evict the cache for that instance. The cache key is `${provider.id}:${instance.id}`, so a helper keyed by `instanceId` (delete keys ending in `:${instanceId}`) is cleaner than requiring the providerId — the instance may not carry it at that point.
2. Also clear on `POST /:id/restart` (operators expect "restart" to re-read config).
3. Longer-term: stop baking mutable instance config into the cached provider — read `enableAutoSplit` (and friends) per-dispatch from the fresh `instance` row, the way `getSplitDelayConfig(instance)` already does on the send path. That removes the whole class of staleness rather than patching each invalidation site.

Option 3 is the real fix; 1+2 are the cheap stopgaps.

## Environment

- Prod: `omni.hv-cx.khal.ai`, omni-api (agno schema), hermes channel.
- Observed 2026-08-14.

vasconceloscezar:
Fixes #906

## Problem

Instance config that gets baked into the agent provider at construction (`enableAutoSplit`, `agentTimeout`, `agentPrefixSenderName`, the resolved `agentId`/`agentType`) had **no effect until a pod restart**. `resolveProvider()` caches the built `IAgentProvider` keyed by `${providerId}:${instanceId}`, and that cache was only invalidated by provider PATCH/DELETE — never by instance mutations, agent-entity mutations, or `POST /instances/:id/restart`. The API returned 200 and the DB showed the new value, but the dispatcher kept the stale provider.

## Fix

- **`agent-dispatcher.ts`** — new `invalidateProviderCacheForInstance(instanceId)`: evicts `${providerId}:${instanceId}` entries by key suffix and disposes them. Unlike `invalidateProviderCache`, it leaves the provider-level shared OpenClaw WS pool up, since instance mutations can't affect connection-level config.
- **`InstanceService.update()`** — evicts when the patch touches a provider-baked field (`agentId`, `agentTimeout`, `enableAutoSplit`, `agentPrefixSenderName`). Presence-guarded so profile/status churn doesn't drop providers holding live NATS/WS subscriptions. Service-level, so REST PATCH, tRPC, and follow-up config paths are all covered. `delete()` now also disposes the cached provider.
- **`AgentService.update()`** — agent-entity fields baked in via `applyAgentFkOverrides` (`name`, `agentProviderId`, `agentType`, `metadata`, `configPath`) evict cached providers for every instance referencing the agent, through both the instance FK and agent routes.
- **`POST /instances/:id/restart`** — evicts too, so restart re-reads config as operators expect (previously it only recycled the channel connection).
- Registered the two new scoped reads in the tenancy db-access guard as `tenant-boundary` (all callers are HTTP routes).

## Tests

New `provider-cache-instance-invalidation.test.ts` pins: per-(provider, instance) caching, targeted eviction leaving sibling instances cached, no-op on unknown ids, no suffix-collision eviction, and the field-presence guard. Full API suite + pre-push gate green.

## Not addressed (pre-existing, noted in #906 as option 3)

The long-term fix — reading mutable instance config per-dispatch instead of baking it into the cached provider — is untouched; this PR closes every mutation path that can leave the cache stale.

### DIFF
--- a/packages/api/src/plugins/agent-dispatcher.ts
+++ b/packages/api/src/plugins/agent-dispatcher.ts
@@ -4514,6 +4514,41 @@ export function invalidateProviderCache(providerId: string): void {
 }
 
 /**
+ * Invalidate cached IAgentProvider instances for a specific instance.
+ *
+ * Instance-level config (enableAutoSplit, agentType, agentTimeout,
+ * agentPrefixSenderName, the resolved agentId) is baked into the provider at
+ * construction time (see createAgnoProvider and friends), so an instance
+ * update must evict the cached entry or the change stays invisible until a
+ * process restart (omni#906). Cache keys are `${providerId}:${instanceId}`;
+ * matching on the `:instanceId` suffix lets callers invalidate without
+ * knowing which provider the instance resolves to.
+ *
+ * Unlike invalidateProviderCache, this leaves the shared OpenClaw client
+ * pool alone — that pool holds provider-level connection state, which an
+ * instance mutation cannot affect.
+ */
+export function invalidateProviderCacheForInstance(instanceId: string): void {
+  const suffix = `:${instanceId}`;
+  let removed = 0;
+
+  for (const [key, provider] of providerCache.entries()) {
+    if (!key.endsWith(suffix)) continue;
+    providerCache.delete(key);
+    removed++;
+    if (provider.dispose) {
+      provider.dispose().catch((err) => {
+        log.warn('Error disposing provider on instance cache invalidation', { key, error: String(err) });
+      });
+    }
+  }
+
+  if (removed > 0) {
+    log.debug('Provider cache invalidated for instance', { instanceId, removed });
+  }
+}
+
+/**
  * Look up provider from DB and resolve to IAgentProvider.
  * Accepts DispatchInstance which carries the transient agentProviderId stamped by applyAgentFkOverrides.
  */
--- a/packages/api/src/plugins/agent-dispatcher.ts
+++ b/packages/api/src/plugins/agent-dispatcher.ts
@@ -4524,9 +4524,23 @@ export function invalidateProviderCache(providerId: string): void {
  * matching on the `:instanceId` suffix lets callers invalidate without
  * knowing which provider the instance resolves to.
  *
- * Unlike invalidateProviderCache, this leaves the shared OpenClaw client
- * pool alone — that pool holds provider-level connection state, which an
- * instance mutation cannot affect.
+ * Unlike invalidateProviderCache, this never touches the shared OpenClaw
+ * client pool — that pool holds provider-level connection state, which an
+ * instance mutation cannot affect — and accordingly it does NOT dispose
+ * evicted OpenClaw providers: their dispose() stops the pooled client that
+ * every sibling instance of the provider is dispatching through, and nothing
+ * on this path would restart or replace the pool entry. Dropping the wrapper
+ * is the whole eviction there. Every other schema owns its resources (e.g.
+ * the NATS Genie per-provider connection and reply subscription) and must be
+ * disposed, or the rebuilt provider would subscribe alongside the leaked one
+ * and every reply would be delivered twice.
+ *
+ * GRANULARITY CAVEAT: the cache key is route-blind. When agent routes on one
+ * instance target different agents of the SAME provider, they share this one
+ * entry, so eviction guarantees a rebuild from fresh rows but whichever route
+ * dispatches first decides which agent's config gets baked in — a pre-existing
+ * limit of the `${providerId}:${instanceId}` key, not something per-instance
+ * eviction can fix.
  */
 export function invalidateProviderCacheForInstance(instanceId: string): void {
   const suffix = `:${instanceId}`;
@@ -4536,6 +4550,7 @@ export function invalidateProviderCacheForInstance(instanceId: string): void {
     if (!key.endsWith(suffix)) continue;
     providerCache.delete(key);
     removed++;
+    if (provider.schema === 'openclaw') continue;
     if (provider.dispose) {
       provider.dispose().catch((err) => {
         log.warn('Error disposing provider on instance cache invalidation', { key, error: String(err) });
--- a/packages/api/src/services/instances.ts
+++ b/packages/api/src/services/instances.ts
@@ -44,6 +44,7 @@ import { NotFoundError } from '@omni/core';
 import type { Database } from '@omni/db';
 import { type ChannelType, type Instance, type NewInstance, instances } from '@omni/db';
 import { and, eq, inArray, sql } from 'drizzle-orm';
+import { invalidateProviderCacheForInstance } from '../plugins/agent-dispatcher';
 import { forgetInstanceOwner, rememberInstanceOwners } from '../tenancy/instance-owner-registry';
 import { credentialSealingEngages, openCredentialField, sealCredentialField } from '../tenancy/sealed-credentials';
 import { currentTenantScope, scopedHandle } from '../tenancy/tenant-scope';
@@ -130,6 +131,25 @@ function openInstanceCredentials<T extends { tenantId?: string | null }>(row: T)
 /** Open a batch of loaded rows. Identity when nothing in the batch is sealed. */
 function openInstanceCredentialsAll<T extends { tenantId?: string | null }>(rows: T[]): T[] {
   return rows.map(openInstanceCredentials);
+}
+
+/**
+ * The `instances` columns that get baked into a cached IAgentProvider at
+ * construction time (see resolveProvider / createAgnoProvider and friends in
+ * agent-dispatcher). An update touching any of these must evict the instance's
+ * provider-cache entry or the running dispatcher keeps the stale values until
+ * the process restarts (omni#906).
+ */
+const PROVIDER_BAKED_COLUMNS = [
+  'agentId',
+  'agentTimeout',
+  'enableAutoSplit',
+  'agentPrefixSenderName',
+] as const satisfies readonly (keyof NewInstance)[];
+
+/** Presence check, not value diff: `null` is a meaningful new value here. */
+export function touchesProviderBakedConfig(data: Partial<NewInstance>): boolean {
+  return PROVIDER_BAKED_COLUMNS.some((column) => column in data);
 }
 
 export class InstanceService {
@@ -351,6 +371,15 @@ export class InstanceService {
       throw new NotFoundError('Instance', id);
     }
 
+    // omni#906: these fields are baked into the cached IAgentProvider at
+    // construction, so the dispatcher would keep serving the old values until
+    // a process restart unless the cache entry is evicted here. Guarded by
+    // field presence so unrelated updates (profileName, isActive, presence…)
+    // don't churn providers that hold live NATS/WS subscriptions.
+    if (touchesProviderBakedConfig(data)) {
+      invalidateProviderCacheForInstance(id);
+    }
+
     rememberInstanceOwners([updated]);
     return openInstanceCredentials(updated);
   }
@@ -400,6 +429,10 @@ export class InstanceService {
     // The instance is gone; drop its ownership entry so the registry stays
     // bounded by what actually exists.
     forgetInstanceOwner(id);
+
+    // Dispose the cached agent provider (and its NATS/WS subscriptions) —
+    // nothing can dispatch for this instance anymore (omni#906).
+    invalidateProviderCacheForInstance(id);
 
     if (this.eventBus) {
       await this.eventBus.publish('instance.disconnected', {
--- a/packages/api/src/services/instances.ts
+++ b/packages/api/src/services/instances.ts
@@ -47,7 +47,7 @@ import { and, eq, inArray, sql } from 'drizzle-orm';
 import { invalidateProviderCacheForInstance } from '../plugins/agent-dispatcher';
 import { forgetInstanceOwner, rememberInstanceOwners } from '../tenancy/instance-owner-registry';
 import { credentialSealingEngages, openCredentialField, sealCredentialField } from '../tenancy/sealed-credentials';
-import { currentTenantScope, scopedHandle } from '../tenancy/tenant-scope';
+import { currentTenantScope, runAfterTenantCommit, scopedHandle } from '../tenancy/tenant-scope';
 
 export interface ListInstancesOptions {
   channel?: ChannelType[];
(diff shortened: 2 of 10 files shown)


vasconceloscezar (commented):
## Code review

The core idea is right and the baked-column list is complete — I checked every `create*Provider` factory: the only other instance fields baked in are `channel` and `tenantId`, neither of which `InstanceService.update()` can change. The suffix match is safe (provider/instance ids are UUIDs, so keys hold exactly one `:`), no direct `db.update(instances)` writer bypasses the service for a baked column, the two new db-access-guard entries are correctly classified as `tenant-boundary`, and the `agents.test.ts` mock change is sound.

The problem is that the eviction primitive calls `dispose()` on providers that hold **shared** resources. Three inline comments below; two more findings that land outside the diff:

---

### `packages/api/src/services/routes.ts:156` — agent_routes mutations still leave a stale cached provider

`mergeRouteOverrides` bakes route-level `agentTimeout`, `enableAutoSplit`, `agentPrefixSenderName` and `route.agentId` into the effective instance passed to `resolveProvider`, and the new comment in `services/agents.ts:137` acknowledges that routes stamp onto the same `${providerId}:${instanceId}` entry. But `RouteService.update/create/delete` never call `invalidateProviderCacheForInstance(route.instanceId)`, so PATCHing a route's `agentTimeout` or repointing its `agentId` keeps the stale provider until a pod restart — the same class of bug this PR fixes for instances and agents. Worth either closing here or trimming the "closes every mutation path" claim in the PR body.

### `packages/api/src/plugins/agent-dispatcher.ts:4431` — cache key ignores the route, so eviction rebuilds nondeterministically (pre-existing)

`cacheKey` is `${provider.id}:${instance.id}`, while the cached provider carries route-resolved `agentId`/`agentType`/timeouts. Two routes on one instance targeting different agents on the same provider share one entry: after `AgentService.update()` evicts, the next dispatch — possibly from the *other* route — repopulates the entry, and traffic on the first route is then served with the second route's baked agent id. Not introduced here, but it bounds what this fix can guarantee, so the eviction docstring shouldn't imply per-instance eviction is sufficient for route-scoped agents.

packages/api/src/plugins/agent-dispatcher.ts
```diff
@@ -4513,6 +4513,41 @@ export function invalidateProviderCache(providerId: string): void {
...
+ */
+export function invalidateProviderCacheForInstance(instanceId: string): void {
+  const suffix = `:${instanceId}`;
+  let removed = 0;
+
+  for (const [key, provider] of providerCache.entries()) {
+    if (!key.endsWith(suffix)) continue;
+    providerCache.delete(key);
+    removed++;
+    if (provider.dispose) {
```
vasconceloscezar, line 4554: **HIGH — disposing an OpenClaw provider stops the *shared* pooled WS client, and the stopped client stays in the pool.**

The docstring above says this "leaves the shared OpenClaw client pool alone", but `dispose()` reaches it anyway. `OpenClawAgentProvider.dispose()` (`packages/core/src/providers/openclaw/provider.ts:555`) calls `this.client.stop()`, and `this.client` is the per-(provider, tenant) pooled `OpenClawClient` from `openclawClientPool` — explicitly shared across every instance of that provider ("DEC-3: Reuse the shared WS client per provider"). `stop()` sets `closed = true`, closes the socket, and flushes all pending requests with `Error('Client stopped')`.

Unlike `invalidateProviderCache`, this function deliberately does **not** delete the pool entry, and `createOpenClawProviderInstance` reuses a pooled client with only `if (!client)` — never re-calling `start()`, which is the only thing that clears `closed`.

Concrete scenario: operator PATCHes `enableAutoSplit` on instance A (or hits `POST /instances/A/restart`, or deletes A) on an OpenClaw provider → the shared WS is stopped, in-flight requests for instances A, B and C all reject, and every subsequent `resolveProvider` rebuilds a provider around the same permanently-closed client. OpenClaw dispatch is dead for that provider+tenant until a process restart or an unrelated provider PATCH.

This is a regression: before this PR only a provider PATCH could stop that client, and that path removes it from the pool so the next resolve builds a fresh one.

Fix: skip `dispose()` here, or also `openclawClientPool.delete(...)` / `start()` the client.

packages/api/src/plugins/agent-dispatcher.ts
```diff
@@ -4513,6 +4513,41 @@ export function invalidateProviderCache(providerId: string): void {
...
+      provider.dispose().catch((err) => {
```
vasconceloscezar, line 4555: **MEDIUM — eviction silently drops in-flight NATS Genie replies.**

`NatsGenieProvider.dispose()` unsubscribes `omni.reply.<instanceId>.>` and drains the connection. That subscription is a *core* NATS subscribe with no JetStream replay (`nats-genie-provider.ts:209`), and it is only re-established lazily when the next inbound message rebuilds the provider.

Scenario: a turn is dispatched fire-and-forget; the operator saves the instance config form (or restarts the instance) while the agent is thinking; the reply is published to a subject with no subscriber; the user never gets an answer, and nothing logs an error.

Consider not disposing on config-driven eviction (drop the map entry and let GC/idle handle it), or re-subscribing eagerly.

packages/api/src/services/instances.ts
```diff
@@ -351,6 +371,15 @@ export class InstanceService {
       throw new NotFoundError('Instance', id);
     }
 
+    // omni#906: these fields are baked into the cached IAgentProvider at
+    // construction, so the dispatcher would keep serving the old values until
+    // a process restart unless the cache entry is evicted here. Guarded by
+    // field presence so unrelated updates (profileName, isActive, presence…)
+    // don't churn providers that hold live NATS/WS subscriptions.
+    if (touchesProviderBakedConfig(data)) {
+      invalidateProviderCacheForInstance(id);
```
vasconceloscezar: **MEDIUM — eviction happens before the request's tenant transaction commits, so the stale entry can be re-populated.**

Under a tenant credential the whole request runs inside one transaction (`runInTenantScope`), and this evicts right after the `UPDATE` statement but potentially many milliseconds before commit. A concurrent inbound message on another connection reads the *pre-commit* instance row and re-caches a provider with the old `enableAutoSplit`/`agentTimeout`; the PATCH then commits and the stale provider persists indefinitely — exactly the bug #906 describes. If the request later rolls back, the eviction also disposed a live provider for nothing (see the OpenClaw/NATS comments).

Same shape at `services/agents.ts:142` and `delete()` below (`services/instances.ts:435`), where a concurrent dispatch can even re-create a provider for a row that is about to disappear.

Scheduling the eviction on commit (or at the route boundary, after the response) would close it.

packages/api/src/plugins/agent-dispatcher.ts
vasconceloscezar: Fixed in d62aaf28. `invalidateProviderCacheForInstance` now skips `dispose()` for OpenClaw providers — the eviction loop has `if (provider.schema === 'openclaw') continue;` before the dispose block, so the shared pooled client is never stopped and the pool entry stays valid. Dropping the wrapper from `providerCache` is the whole eviction there: the next dispatch rebuilds an `OpenClawAgentProvider` from fresh instance rows around the same live pooled client (`createOpenClawProviderInstance` reuses it via the `if (!client)` check, which is now safe again since nothing has set `closed`). The docstring was rewritten to state this explicitly instead of the incorrect "leaves the pool alone" claim.

Pinned by a new test in `provider-cache-instance-invalidation.test.ts` ("openclaw eviction — the shared pooled WS client must survive"): it injects a mock client via `__test__.OpenClawClientClass`, resolves a provider, evicts, and asserts `stopCalls === 0`, exactly one client ever constructed, the rebuilt provider is a new object, and `__test__.openclawPoolKeys()` still contains the pool entry.

Every other schema still disposes — see my reply on the NATS Genie comment for why that asymmetry is deliberate.

packages/api/src/plugins/agent-dispatcher.ts
vasconceloscezar: The scenario is real, but I kept `dispose()` for NATS Genie deliberately — the suggested "drop the map entry and let GC/idle handle it" would trade a rare dropped reply for permanent duplicate delivery.

Evidence: `startReplySubscription` subscribes with `this.nc.subscribe(topic)` (`nats-genie-provider.ts:209`) — a plain core-NATS subscribe with **no queue group**, on a **per-provider connection** the provider owns. If the evicted provider is leaked instead of disposed, its subscription on `omni.reply.<instanceId>.>` stays live forever (nothing ever disposes it later), and the moment the next inbound message rebuilds a provider and subscribes, *both* subscriptions receive every reply and `onReply` delivers each answer to the user twice, indefinitely. A reply dropped in the narrow window between a config save and the next inbound message is the lesser harm.

What d62aaf28 does change narrows that window to real cases: eviction is now deferred to transaction commit via `runAfterTenantCommit` and dropped on rollback, so an aborted PATCH no longer tears down the subscription at all, and the field-presence guard in `InstanceService.update()` means only genuinely baked-config changes evict. Eager re-subscribe (make-before-break) would need the dispatcher to re-resolve fresh provider+instance rows outside a dispatch — doable, but it belongs with the issue's option 3 (stop baking config into cached providers) rather than this stopgap.

packages/api/src/services/instances.ts
vasconceloscezar: Fixed in d62aaf28, via a new `runAfterTenantCommit(fn)` in `tenancy/tenant-scope.ts`: `TenantScopeState` grew a `postCommit` callback array, and `runInTenantScope` now awaits `withTenantTransaction` and runs the queued callbacks only after it resolves — i.e. after COMMIT — while a throw (rollback) skips them entirely, so a rolled-back PATCH no longer disposes a live provider. With no scope active the callback runs immediately, which is already post-commit there since the ambient pool autocommits per statement. Callbacks are best-effort (per-callback try/catch), so an eviction failure can't turn a committed request into an error response.

All eviction sites now defer through it: `InstanceService.update()` and `delete()`, `AgentService.update()` (the instance/route lookup still runs inside the transaction; only the eviction waits), and the `RouteService.create/update/delete` evictions added for the route-mutation gap from the review body. Covered by four new tests in `tenant-scope.test.ts` (immediate without scope, ordered after the transaction body, dropped on rollback, best-effort on callback failure).

One honest residual: a dispatch that loaded its instance row just before commit can still call `resolveProvider` just after the post-commit eviction and cache a provider built from that stale read. Closing that completely needs generation-stamped cache entries (evict bumps a counter; `resolveProvider` refuses to cache a build older than the counter). This change shrinks the race from "any concurrent dispatch during the transaction" to that instant; the generation stamp can ride along with the issue's option 3 refactor if it shows up in practice.

PR #936, opened 2026-08-30
feat(webhooks): verify source signatures and lock down generic ingress

### LINKED ISSUE #928: security(webhooks): generic ingress lacks source signature verification + autoCreate of sources on receive
## Context

While auditing the generic webhook ingress path (`POST /api/v2/webhooks/:source`) during the #925 discussion (omni as event backbone), three security/hardening gaps surfaced. None is open injection — the route sits behind `authMiddleware` (`app.ts:456` → `protectedApp.route('/', v2Routes)`), so it requires an API key. But the current design cannot safely support a real external source.

### 1. No source signature verification

`WebhooksService.receive()` (`packages/api/src/services/webhooks.ts`) validates `expectedHeaders` by checking header **presence** only:

```ts
if (!headers[headerName.toLowerCase()]) {
  throw new Error(`Missing required header: ${headerName}`);
}
```

No HMAC verification (GitHub's `X-Hub-Signature-256`, `Stripe-Signature`, ClickUp equivalents). Any authenticated request with the header present — with ANY value — passes. A `webhook-auth.ts` middleware that does secret verification exists, but it's Telegram-specific (`X-Telegram-Bot-Api-Secret-Token`) and doesn't cover the generic receiver.

### 2. `autoCreate: true` on the receiver

The route (`packages/api/src/routes/v2/webhooks.ts:124`) calls `receive(..., { autoCreate: true })`: a POST to a **nonexistent** `:source` **auto-creates the source** and publishes the event to the bus. Any API key holder can materialize new sources and inject `custom.webhook.{arbitrary-name}` events that automations may consume — with no prior administrative act.

### 3. Credential model incompatible with third parties

Since the route requires an omni API key, plugging in a real external source (GitHub, Stripe, ClickUp) means embedding one of our API keys — long-lived, API-scoped — into a third party's webhook config. If it leaks there, it's access to our API. The industry pattern is the inverse: public endpoint + per-request signature with a source-dedicated secret.

## Proposal

1. **`signature_config` per webhook source** (`webhook_sources` gains: secret ref, algorithm `hmac-sha256`/`hmac-sha1`/`token-match`, header name). Verified BEFORE publishing; failure → 401, nothing enters the journal. Generalizes the pattern `webhook-auth.ts` already applies for Telegram.
2. **`autoCreate` off by default** — sources are created by administrative act (API/CLI), explicit flag for dev environments.
3. **Dedicated auth-exempt ingress route** `POST /webhooks/ingress/:source` (outside `protectedApp`), protected exclusively by the source signature — following the precedent of the channel webhooks (Telegram/Gupshup/Meta), which are already auth-exempt with their own verification. The current authenticated route can remain for internal use.
4. Rate limiting on the public endpoint (same treatment as channel webhooks).

## Severity

Medium today (requires a valid API key; surface = key holders + the scope of automations consuming `custom.webhook.*`). **Becomes high the moment the first real external source is plugged in** — which is exactly Phase 3 of #925. Therefore: prerequisite for Phase 3, blocking for any GitHub/ClickUp/Stripe integration via the generic webhook.

## Related

- #925 (event backbone RFC) — this hardening was the RFC's "G0", extracted into its own issue.

vasconceloscezar:
## Summary

Implements the #928 hardening for the generic webhook receiver (`POST /api/v2/webhooks/:source`). Closes #928.

### 1. Per-source signature verification
- `webhook_sources` gains `signature_config` (`hmac-sha256` / `hmac-sha1` / `token-match` + header name + optional digest prefix like `sha256=`) and `signature_secret`, sealed per-tenant via `sealCredentialField` like other credential columns (migration `0052`, hand-written per the 0044–0051 precedent, additive + idempotent).
- `WebhookService.receive` verifies the signature **over the raw request body** with a timing-safe compare **before** the stats update and publish — a failed verification is a typed 401 and nothing enters the journal.
- The secret is write-only: CRUD responses return `hasSignatureSecret` instead of the value.

### 2. autoCreate off by default
- `receive()` now defaults `autoCreate: false`; sources are created by administrative act (API/CLI). `OMNI_WEBHOOK_AUTOCREATE=true` restores the old behavior for dev environments. **Behavioral change:** an authenticated POST to an unknown source is now 404 instead of silently materializing it.

### 3. Dedicated auth-exempt ingress route
- `POST /api/v2/webhooks/ingress/:source`, mounted before `protectedApp` following the channel-webhook precedent (Telegram/Meta/Twilio). Reachable **only** for sources with a signature config; never auto-creates; registered in the route-ownership gate as public-by-contract.
- All rejections (unknown source, disabled, unconfigured, bad signature) collapse into one 401 shape so the endpoint is not a source-name existence oracle; the real reason is logged server-side.
- The authenticated `/webhooks/:source` route remains for internal use (and now also verifies signatures when configured).

### 4. Rate limiting
- The public ingress gets its own IP-keyed limiter (events budget, isolated key prefix).

### Tooling
- SDK: `WebhookSignatureConfigBody` + new fields on `CreateWebhookSourceBody`.
- CLI: `omni webhooks create/update --signature-algorithm --signature-header --signature-prefix --signature-secret`, plus `--clear-signature` on update.
- OpenAPI: source schemas updated, ingress path registered with `security: []`.

> Note: generated SDKs (`packages/sdk*/generated`) were **not** regenerated — `make sdk-generate` needs a running API. Follow-up on next spec regen.

## Testing
- `bun test` full suite via pre-push gate: 5831 pass / 0 fail.
- New app-level tests (`webhook-ingress.test.ts`): signed request accepted with no API key; bad signature / unknown source / unconfigured source / disabled source all return the uniform 401; authenticated route still 401s without a key.
- New service tests: HMAC valid/invalid, token-match, missing header, missing raw body, missing secret, autoCreate-off default, secret invariants on create/update.
- `make typecheck` and `make lint` green.

### DIFF
--- a/packages/api/src/app.ts
+++ b/packages/api/src/app.ts
@@ -443,17 +443,26 @@ export function createApp(
   // unconfigured, bad signature) collapse into one 401 shape so the public
   // endpoint is not a source-name existence oracle; the real reason is logged.
   // The authenticated POST /api/v2/webhooks/:source route remains for internal
-  // callers. Must be mounted before protectedApp.
+  // callers. A non-empty body that is not a JSON object is a 400 — silently
+  // publishing an empty payload would fire automations on a hollow event.
+  // Must be mounted before protectedApp.
   app.post('/api/v2/webhooks/ingress/:source', webhookIngressRateLimitMiddleware, async (c) => {
     const sourceName = c.req.param('source');
     const services = c.get('services');
 
     const rawBody = await c.req.text();
-    let payload: Record<string, unknown>;
-    try {
-      payload = rawBody ? JSON.parse(rawBody) : {};
-    } catch {
-      payload = {};
+    let payload: Record<string, unknown> = {};
+    if (rawBody) {
+      let parsed: unknown;
+      try {
+        parsed = JSON.parse(rawBody);
+      } catch {
+        parsed = undefined;
+      }
+      if (typeof parsed !== 'object' || parsed === null || Array.isArray(parsed)) {
+        return c.json({ error: { code: 'VALIDATION', message: 'Request body must be a JSON object' } }, 400);
+      }
+      payload = parsed as Record<string, unknown>;
     }
 
     const headers: Record<string, string> = {};
--- a/packages/api/src/app.ts
+++ b/packages/api/src/app.ts
@@ -6,7 +6,7 @@ import { existsSync } from 'node:fs';
 import path from 'node:path';
 import type { ChannelRegistry } from '@omni/channel-sdk';
 import type { WhatsAppBusinessPlugin } from '@omni/channel-whatsapp-business';
-import { type EventBus, createLogger } from '@omni/core';
+import { type EventBus, NotFoundError, OmniError, createLogger } from '@omni/core';
 import { type Database, type Instance, whatsappFlowKeys } from '@omni/db';
 import { eq } from 'drizzle-orm';
 import { Hono } from 'hono';
@@ -76,7 +76,7 @@ import { tenancyMiddleware } from './middleware/tenancy';
 
 import { createContextMiddleware } from './middleware/context';
 import { errorHandler } from './middleware/error';
-import { rateLimitMiddleware } from './middleware/rate-limit';
+import { rateLimitMiddleware, webhookIngressRateLimitMiddleware } from './middleware/rate-limit';
 import { defaultTimeoutMiddleware } from './middleware/timeout';
 import { versionHeadersMiddleware } from './middleware/version-headers';
 import { createWebhookAuthMiddleware } from './middleware/webhook-auth';
@@ -432,6 +432,52 @@ export function createApp(
     }
 
     return plugin.handleWebhook(c.req.raw);
+  });
+
+  // Public generic webhook ingress — auth-exempt, verified EXCLUSIVELY by the
+  // source's signature config (issue #928), following the channel-webhook
+  // precedent (Telegram/Meta/Twilio above). A source is reachable here only
+  // when an admin has configured `signatureConfig` + secret on it; verification
+  // runs over the raw body BEFORE anything is published. Sources are never
+  // auto-created on this surface. All rejections (unknown source, disabled,
+  // unconfigured, bad signature) collapse into one 401 shape so the public
+  // endpoint is not a source-name existence oracle; the real reason is logged.
+  // The authenticated POST /api/v2/webhooks/:source route remains for internal
+  // callers. Must be mounted before protectedApp.
+  app.post('/api/v2/webhooks/ingress/:source', webhookIngressRateLimitMiddleware, async (c) => {
+    const sourceName = c.req.param('source');
+    const services = c.get('services');
+
+    const rawBody = await c.req.text();
+    let payload: Record<string, unknown>;
+    try {
+      payload = rawBody ? JSON.parse(rawBody) : {};
+    } catch {
+      payload = {};
+    }
+
+    const headers: Record<string, string> = {};
+    for (const [key, value] of Object.entries(c.req.header())) {
+      headers[key.toLowerCase()] = value ?? '';
+    }
+
+    try {
+      const result = await services.webhooks.receive(sourceName, payload, headers, {
+        autoCreate: false,
+        rawBody,
+        requireSignature: true,
+      });
+      return c.json(result);
+    } catch (error) {
+      const rejected =
+        error instanceof NotFoundError ||
+        (error instanceof OmniError && (error.code === 'UNAUTHORIZED' || error.code === 'FORBIDDEN'));
+      if (rejected) {
+        httpLog.warn('Webhook ingress rejected', { sourceName, reason: (error as Error).message });
+        return c.json({ error: { code: 'UNAUTHORIZED', message: 'Webhook verification failed' } }, 401);
+      }
+      throw error;
+    }
   });
 
   // ── Multitenancy control plane — feature-flagged, OFF by default ────────────
--- a/packages/api/src/middleware/rate-limit.ts
+++ b/packages/api/src/middleware/rate-limit.ts
@@ -42,9 +42,13 @@ const RATE_LIMITS = {
 } as const;
 
 /**
- * Create rate limiting middleware
+ * Create rate limiting middleware.
+ *
+ * `keyPrefix` isolates a limiter's counters from the default bucket so a
+ * surface with its own budget (e.g. the public webhook ingress) doesn't share
+ * windows with the general limiter for the same identifier.
  */
-function createRateLimiter(config: RateLimitConfig = RATE_LIMITS.general) {
+function createRateLimiter(config: RateLimitConfig = RATE_LIMITS.general, keyPrefix = 'ratelimit') {
   return createMiddleware<{ Variables: AppVariables }>(async (c, next) => {
     // Use API key ID as identifier, then only trusted infra-provided address data.
     // Do not trust client-supplied IP headers unless explicitly configured.
@@ -56,7 +60,7 @@ function createRateLimiter(config: RateLimitConfig = RATE_LIMITS.general) {
     const normalizedIp = rawIp?.split(',')[0]?.trim() || undefined;
     const identifier = apiKey?.id ? `api:${apiKey.id}` : normalizedIp ? `ip:${normalizedIp}` : 'anon';
 
-    const key = `ratelimit:${identifier}`;
+    const key = `${keyPrefix}:${identifier}`;
     const now = Date.now();
 
     let entry = rateLimitStore.get(key);
@@ -111,3 +115,9 @@ function createRateLimiter(config: RateLimitConfig = RATE_LIMITS.general) {
  * Default rate limiter for general endpoints
  */
 export const rateLimitMiddleware = createRateLimiter(RATE_LIMITS.general);
+
+/**
+ * Rate limiter for the auth-exempt generic webhook ingress (issue #928).
+ * Keyed by IP (no API key on that surface), same budget as channel events.
+ */
+export const webhookIngressRateLimitMiddleware = createRateLimiter(RATE_LIMITS.events, 'ratelimit:webhook-ingress');
--- a/packages/api/src/middleware/rate-limit.ts
+++ b/packages/api/src/middleware/rate-limit.ts
@@ -55,7 +55,14 @@ function createRateLimiter(config: RateLimitConfig = RATE_LIMITS.general, keyPre
     const apiKey = c.get('apiKey');
     const trustedProxyHeader = process.env.TRUSTED_PROXY_HEADER?.toLowerCase();
     const trustedProxyIp = trustedProxyHeader ? c.req.header(trustedProxyHeader) : undefined;
-    const remoteAddr = (c.env as { remoteAddr?: string } | undefined)?.remoteAddr;
+    // Under Bun.serve, Hono's env is the Bun Server, which exposes the peer
+    // address via requestIP() — there is no `remoteAddr` field. Without this,
+    // unauthenticated surfaces (the public webhook ingress) would all share
+    // one 'anon' bucket instead of being limited per IP.
+    const server = c.env as
+      | { remoteAddr?: string; requestIP?: (req: Request) => { address?: string } | null }
+      | undefined;
+    const remoteAddr = server?.remoteAddr ?? server?.requestIP?.(c.req.raw)?.address;
     const rawIp = trustedProxyIp ?? remoteAddr;
     const normalizedIp = rawIp?.split(',')[0]?.trim() || undefined;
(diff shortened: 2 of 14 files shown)


packages/api/src/middleware/rate-limit.ts
```diff
@@ -111,3 +115,9 @@ function createRateLimiter(config: RateLimitConfig = RATE_LIMITS.general) {
  * Default rate limiter for general endpoints
  */
 export const rateLimitMiddleware = createRateLimiter(RATE_LIMITS.general);
+
+/**
+ * Rate limiter for the auth-exempt generic webhook ingress (issue #928).
+ * Keyed by IP (no API key on that surface), same budget as channel events.
+ */
+export const webhookIngressRateLimitMiddleware = createRateLimiter(RATE_LIMITS.events, 'ratelimit:webhook-ingress');
```
vasconceloscezar, line 130: **medium — the ingress limiter degenerates to one global bucket, not per-IP**

`createRateLimiter` derives its identifier from `apiKey?.id` -> `TRUSTED_PROXY_HEADER` -> `c.env.remoteAddr`. On this auth-exempt route there is no API key; `TRUSTED_PROXY_HEADER` appears nowhere in the repo outside line 56 (no env example, no helm values, no docs), and `c.env.remoteAddr` is never populated — nothing in `packages/api` assigns it, and Hono's Bun adapter exposes `requestIP()` rather than that field.

So the identifier is the constant `'anon'` and the key is `ratelimit:webhook-ingress:anon` for every caller. One unauthenticated client sending 100 req/min exhausts the shared window and every legitimate provider (GitHub, Stripe, ...) gets 429 for the rest of the minute; normal multi-source traffic will also self-throttle.

The route-ownership justification at `route-ownership.ts:279` claims "the surface is IP-rate-limited", which is not what happens in the default deployment.

packages/api/src/services/webhooks.ts
```diff
@@ -86,9 +118,30 @@ export class WebhookService {
    * Update a webhook source
    */
   async update(id: string, data: Partial<NewWebhookSource>): Promise<WebhookSource> {
+    const patch: Partial<NewWebhookSource> = { ...data };
+
+    if (data.signatureSecret != null) {
```
vasconceloscezar: **medium — `PATCH { signatureSecret: null }` silently breaks a configured source**

The route schema allows `signatureSecret: null` on its own. Here `data.signatureSecret != null` is false, and the `else if` requires `data.signatureConfig === null` — so neither branch runs, but `patch` still carries `signatureSecret: null` from the spread. The secret is nulled while `signature_config` stays set. The `if (data.signatureConfig)` guard below doesn't fire either (config is `undefined`), so no `ValidationError`.

The admin gets a 200 with `hasSignatureSecret: false`, and from then on every ingress delivery for that source hits the `if (!secret)` path in `verifySignature` and returns the uniform 401 — the config => secret invariant the surrounding code enforces is bypassed on exactly this path.

Same shape in the mirror direction: `{ signatureConfig: null, signatureSecret: "new" }` seals a new secret while clearing the config, orphaning it.

packages/api/src/services/webhooks.ts
```diff
@@ -137,10 +195,29 @@ export class WebhookService {
...
+      });
+    }
+
+    if (source.signatureConfig) {
+      this.verifySignature(source, rawBody, headers);
+    }
+
+    // Validate expected headers if configured (presence only — the signature
+    // check above is the authenticity gate)
     if (source.expectedHeaders) {
```
vasconceloscezar, line 230: **low — missing `expectedHeaders` yields a 500 on the public ingress, not the uniform 401**

The `throw new Error(...)` on line 224 is neither a `NotFoundError` nor an `OmniError`, so the `rejected` test at `app.ts:472` is false and it rethrows into `errorHandler` -> 500 `INTERNAL_ERROR`.

A source configured with both `signatureConfig` and `expectedHeaders` that receives a correctly signed request missing one of those headers gets an opaque 500 (and a Sentry event) instead of the documented 4xx. It also punches a hole in the "one 401 shape" contract the route comment and the OpenAPI description assert.

packages/api/src/app.ts
```diff
@@ -434,6 +434,52 @@ export function createApp(
...
+  // endpoint is not a source-name existence oracle; the real reason is logged.
+  // The authenticated POST /api/v2/webhooks/:source route remains for internal
+  // callers. Must be mounted before protectedApp.
+  app.post('/api/v2/webhooks/ingress/:source', webhookIngressRateLimitMiddleware, async (c) => {
+    const sourceName = c.req.param('source');
+    const services = c.get('services');
+
+    const rawBody = await c.req.text();
+    let payload: Record<string, unknown>;
+    try {
```
vasconceloscezar: **low — a validly signed non-JSON body is acked 200 with an empty payload**

`JSON.parse` failure falls back to `payload = {}`, verification still succeeds (it runs over `rawBody`), and the handler publishes `custom.webhook.<source>` with only `{ source }` and returns 200.

A provider posting `application/x-www-form-urlencoded` (common — Twilio-style senders) or any malformed JSON sees success while the payload is silently dropped and downstream automations fire on an empty event. Rejecting with a 4xx, or gating on Content-Type, would make the loss visible.

packages/api/src/services/webhooks.ts
```diff
@@ -186,6 +263,46 @@ export class WebhookService {
     };
   }
 
+  /**
```
vasconceloscezar, line 275: **low — no replay protection on the auth-exempt surface**

The signature covers the body only: no timestamp/nonce tolerance, no delivery-id dedupe. Anyone who observes one signed request (a proxy, a log, a mirrored delivery) can replay it unbounded against `/api/v2/webhooks/ingress/:source`, re-publishing the event every time — and these events drive automations.

Worth at least documenting as an accepted limitation in the route-ownership justification, which currently reads as if the signature alone is sufficient.

packages/api/src/services/webhooks.ts
```diff
@@ -186,6 +263,46 @@ export class WebhookService {
...
+    }
+
+    const secret = openCredentialField(source.tenantId ?? this.tenantId, source.signatureSecret);
+    if (!secret) {
+      log.warn('Webhook source has a signature config but no usable secret', { sourceName });
+      throw unauthorized('Webhook signature verification unavailable for this source');
+    }
+
+    let expected: string;
+    if (config.algorithm === 'token-match') {
```
vasconceloscezar, line 300: **low — `prefix` is silently ignored for `token-match`**

`expected = secret` never consults `config.prefix`, but nothing in the schema, the CLI (`--signature-prefix` is offered alongside `--signature-algorithm token-match`), or the OpenAPI description forbids the combination.

An admin who configures `algorithm: 'token-match'` with a prefix gets a source that rejects every delivery with "Invalid webhook signature" and no hint why. Either reject the combination at validation time or honor the prefix.

packages/cli/src/commands/webhooks.ts
```diff
@@ -74,36 +95,55 @@ export function createWebhooksCommand(): Command {
...
+            enabled: !options.disabled,
+            expectedHeaders,
+            signatureConfig,
+            signatureSecret: options.signatureSecret,
+          });
+
+          output.success(`Webhook source created: ${source.id}`, {
+            id: source.id,
+            name: source.name,
+            url: `POST /api/v2/webhooks/${source.id}`,
```
vasconceloscezar: **low — the URL printed after `webhooks create` is wrong, and now doubly so**

This prints `POST /api/v2/webhooks/${source.id}`, but both receivers key on the source **name** (`getByName`), not the id — so the printed URL 404s. After this PR the URL an external sender actually needs is `POST /api/v2/webhooks/ingress/<name>`.

Since the command now grows signature flags whose whole purpose is enabling that public route, this is the line users will copy.

vasconceloscezar (commented):
Automated code review of the full diff (`main...HEAD`). 7 findings: 2 medium, 5 low. Findings 1 and 2 were independently re-verified against the tree; the rest are single-pass. New tests pass (43 + 36).

packages/api/src/middleware/rate-limit.ts
vasconceloscezar: Partially fixed in e2a6fad5. Confirmed the analysis: under `Bun.serve`, Hono's env is the Bun `Server` (`app.fetch(req, server)` in `index.ts`), which exposes the peer via `requestIP()` — nothing ever populated `remoteAddr`, so the identifier did collapse to `'anon'` and the ingress shared one global bucket. `createRateLimiter` now resolves `server.requestIP(c.req.raw)?.address` (API-key identity and `TRUSTED_PROXY_HEADER` precedence unchanged), so a directly-exposed deployment buckets per IP.

Caveat, stated plainly: `TRUSTED_PROXY_HEADER` is still unconfigured in helm/env examples, so in a deployment behind an ingress/LB, `requestIP()` returns the ingress pod's IP — every external caller still lands in one bucket there. The fix closes the finding for direct exposure only; fully closing it as deployed needs `TRUSTED_PROXY_HEADER` set to the ingress's forwarded-for header in the deployment config, which is an ops change I did not make in this PR.

packages/api/src/services/webhooks.ts
vasconceloscezar: Fixed in e2a6fad5. `update()` now materializes the resulting (config, secret) pair for any partial combination: `nextConfig`/`nextSecret` inherit the stored row for untouched fields, then the invariant is enforced both ways — `PATCH {signatureSecret: null}` against a configured source throws `ValidationError('signatureSecret is required when signatureConfig is set')` (400) instead of silently breaking delivery, and `{signatureConfig: null, signatureSecret: 'new'}` throws `ValidationError('signatureSecret cannot be set without a signatureConfig')` instead of sealing an orphan. Clearing the config alone still clears the secret, and `create()` symmetrically rejects a secret without a config. Service tests cover all four quadrants (webhooks.test.ts, 'signature secret invariants').

packages/api/src/services/webhooks.ts
vasconceloscezar: Fixed in e2a6fad5 — the expectedHeaders presence check now throws `ValidationError` instead of a bare `Error`, so it maps to 400 VALIDATION through the global error handler instead of a 500 + Sentry event. One deliberate nuance: it stays a descriptive 400 rather than folding into the uniform 401, because this check runs after `verifySignature` — the caller has already proven possession of the source secret, so the existence-oracle rationale behind the opaque 401 doesn't apply and a diagnosable error is more useful to the legitimate sender. Added an app-level test for the signed-but-missing-header path (webhook-ingress.test.ts).

packages/api/src/app.ts
vasconceloscezar: Fixed in e2a6fad5. The ingress handler in app.ts now rejects a non-empty body that doesn't parse into a JSON object (parse failure, arrays, scalars) with 400 `{code: 'VALIDATION', message: 'Request body must be a JSON object'}` instead of acking 200 with a hollow payload. An empty body still maps to `{}`, since some providers send body-less pings. The authenticated internal `/webhooks/:source` route keeps its historical lenient fallback deliberately — changing it would break existing internal callers, and that surface sits behind an API key. Test added: 'a signed non-JSON body is rejected with 400, not silently emptied'.

packages/api/src/services/webhooks.ts
vasconceloscezar: Documented in e2a6fad5, per your suggested minimum: the route-ownership justification now carries an explicit ACCEPTED LIMITATION — the signature covers the body only, no timestamp tolerance or delivery-id dedupe, so a captured signed request can be replayed within the rate limit, and consumers of `custom.webhook.*` must treat deliveries as at-least-once (which they already must, since upstream providers retry deliveries themselves). I left enforcement out of this PR on purpose: the generic scheme mirrors GitHub's `X-Hub-Signature-256`, which has no timestamp either; real replay resistance (Stripe-style `t=...,v1=...`) needs per-provider verification schemes and belongs in a follow-up rather than bolted onto the generic contract.

packages/api/src/services/webhooks.ts
vasconceloscezar: Fixed in e2a6fad5 — the combination is rejected at every boundary: `signatureConfigSchema` in routes/v2/webhooks.ts refines `token-match` + `prefix` into a 400 ("prefix is not applicable to algorithm 'token-match'") on both create and PATCH (verified the refine survives `.partial()`), `buildSignatureConfig` in the CLI errors on `--signature-prefix` with `token-match`, and the OpenAPI description now marks prefix as HMAC-only. A misconfigured always-rejecting source can no longer be created.

packages/cli/src/commands/webhooks.ts
vasconceloscezar: Fixed in e2a6fad5 — and good catch that the id-vs-name bug predates this PR. `webhooks create` now prints `url: POST /api/v2/webhooks/<name>` (both receivers key on the source NAME via `getByName`), and when signature flags were provided it additionally prints `publicUrl: POST /api/v2/webhooks/ingress/<name>` — the line an external sender actually needs to copy.

PR #937, opened 2026-08-30
feat(slack): migrate to Agent messaging experience (agent_view + Agent Sessions API)

### LINKED ISSUE #914: Slack: migrar para a Agent messaging experience (agent_view + Agent Sessions API) — assistant_view deprecada em fev/2027
## Contexto

O Slack anunciou a **Agent messaging experience** (`agent_view`) e marcou a **`assistant_view` para deprecação em fevereiro de 2027** ([changelog 20/08/2026](https://docs.slack.dev/changelog/2026/08/20/agent-updates/), [anúncio 30/06/2026](https://docs.slack.dev/changelog/2026/06/30/agent-messages-tab/)). Hoje existe uma ponte de compatibilidade, então nada quebrou ainda — mas o painel de qualquer app novo já mostra o aviso *"This app is using an outdated experience"*.

O canal Slack do omni implementa a experiência antiga.

## Evidência (omni `2.260706.1`, `dist/server/index.js`)

| símbolo | ocorrências |
|---|---|
| `assistant.threads.setStatus` | 6 |
| `assistant_thread_started` | 2 |
| `agents.sessions.setStatus` | **0** |
| `agents.sessions.rename` | **0** |
| `agent_session_stopped` | **0** |
| `agent_view` | **0** |

## O que precisa mudar

**1. Agent Sessions API** — `agents.sessions.setStatus` substitui `assistant.threads.setStatus`; `agents.sessions.rename` substitui `assistant.threads.setTitle`.

**2. Evento novo `agent_session_stopped`** — dispara quando o usuário aperta o botão nativo de parar o stream. Hoje o omni não escuta, então **o agente continua rodando depois do usuário mandar parar**. Num provider `claude-code` de tarefa longa isso é dinheiro queimado: no nosso caso um run leva ~18 min e ~US$4,50.

**3. `agent_view` no manifesto** — e o setup do omni deveria documentar/gerar isso. Hoje quem cria um app Slack pelo caminho do omni não recebe nenhuma orientação de habilitar a feature de Agent; o resultado é status nativo que nunca aparece.

---

## Bug relacionado, e independente da migração: status morre em silêncio fora de thread

Em `channel-slack/src/handlers/typing.ts`:

```js
async function setSlackThreadStatus(params) {
  const { client, channelId, threadTs, status, ... } = params;
  if (!threadTs)
    return false;          // <- sai calado, sem log
```

e no dispatcher:

```js
const threadTs = options?.threadId ?? this.activeThreads.get(`${instanceId}:${chatId}`);
if (!threadTs)
  return { delivered: false, method, reason: "no_active_thread" };
```

O `reason: "no_active_thread"` **nunca é logado**. Menção no nível do canal (sem thread) não produz status nem erro — e não há como diagnosticar sem ler o bundle.

**Repro:** instância Slack com provider `claude-code`, mencionar o bot no nível do canal (não em thread), provider levando minutos. Esperado: algum sinal de progresso, ou ao menos um log dizendo por que não houve. Observado: silêncio total, `grep setStatus` no log não retorna nada.

**Sugestão mínima:** `logger.debug`/`info` no caminho do bail, com o `reason`. Custa uma linha e economiza a leitura do bundle.

## Impacto pra quem usa

Rodamos um avaliador de código em cima de `claude-code` no Slack. Numa execução de 18 min o usuário não teve nenhum feedback visual e perguntou várias vezes se estava funcionando. O `reaction-ack` (👀) resolve parcialmente, mas é reação numa mensagem, não o status nativo — que é justamente o que a experiência de Agent do Slack oferece.

## Sugestão de escopo

1. Trocar as chamadas pela Agent Sessions API, mantendo fallback pela ponte de compatibilidade enquanto ela existir
2. Assinar e propagar `agent_session_stopped` → abortar o turno do provider
3. Logar o bail de `no_active_thread`
4. Documentar `agent_view` + escopos no guia de setup do canal Slack

Disposto a contribuir com PR se a direção fizer sentido.

vasconceloscezar:
Closes #914.

Slack deprecated `assistant_view` (removal: February 2027) in favor of the Agent messaging experience. This PR migrates the Slack channel and fixes the related silent-status bug.

## What changed

**1. Agent Sessions API** (`packages/channel-slack/src/handlers/typing.ts`)
- Status calls now prefer `agents.sessions.setStatus` — non-empty status → `processing`, clear → `active`, per Slack's migration guide.
- Per-client fallback to the deprecated `assistant.threads.setStatus` when the Agent Sessions API is unavailable (`unknown_method` / `feature_disabled`), memoized so un-migrated workspaces don't pay a doubled API call per status.
- `assistant.threads.setTitle` → `agents.sessions.rename` was checked: the repo never called setTitle, so there is nothing to migrate there.

**2. Native stop button → provider abort**
- New handler subscribes `agent_session_stopped` (`handlers/agent-sessions.ts`). Subscribing is also what makes Slack *show* the stop button while a session is `processing`.
- On stop, the plugin publishes a new core event `agent.run.cancel_requested` and clears the session status (Slack's contract for the event).
- The agent dispatcher registers a per-run `AbortController` (keyed `instanceId:chatId`, next to `activeStreams`) and feeds `trigger.abortSignal` — already honored end-to-end by the claude-code streaming provider. A cancelled streaming run never falls back to the accumulate path (that would re-run the turn the user just stopped). Accumulate-mode runs that ignore the signal have their reply discarded on completion instead.
- No more 18-minute / ~US$4.50 claude-code runs continuing after the user pressed stop — at least for streaming dispatch; turn-based (fire-and-forget) providers are out of scope here.

**3. Silent status bail is now logged** — both `setSlackThreadStatus` (`no_thread_ts`) and `sendPresenceStatus` (`no_active_thread`) emit a debug line, so a channel-level mention that produces no status is diagnosable without reading the bundle.

**4. Manifest + docs**
- `buildSlackManifest()` emits `features.agent_view` (`agent_description`, optional `suggested_prompts`) and subscribes `agent_session_stopped`.
- New setup guide: `docs/channels/slack.md` (manifest generation, agent_view caveats, stop/status internals).

## Notes for reviewers
- `PresenceStatusResult.method` now reports whichever API actually handled the call; the presence route echoes it (`agents.sessions.setStatus` in the common case).
- `@slack/web-api` 7.15.2 has no typed `agents.*` namespace yet, hence `client.apiCall`.
- The `agent_view` manifest switch is one-way per Slack app; existing apps that apply the new manifest cannot revert to `assistant_view`.

## Testing
- `make typecheck` + `make lint` clean.
- channel-slack: 252 tests pass (new: agent-sessions handler, Agent-Sessions-first status + fallback memoization, manifest agent_view).
- api: dispatcher subscription test updated for the new cancel subscription; full suite green except the pre-existing MinIO-harness timeout (`media-processor-remote.test.ts`, infra-dependent) — unrelated to this change.
- DB-backed presence tests skip locally by design; CI runs them.

### DIFF
--- a/packages/api/src/plugins/agent-dispatcher.ts
+++ b/packages/api/src/plugins/agent-dispatcher.ts
@@ -33,6 +33,7 @@ import type { StreamSender } from '@omni/channel-sdk';
 import {
   A2AAgentProvider,
   AgUiAgentProvider,
+  type AgentRunCancelRequestedPayload,
   type AgentTrigger,
   type AgentTriggerType,
   AgnoAgentProvider,
@@ -2039,6 +2040,46 @@ async function fetchChatMetadata(
 // ─── Per-chatId stream guard ──────────────────────────────
 const activeStreams = new Map<string, StreamSender>();
 
+// ─── Per-chatId run cancellation (#914) ───────────────────
+// Populated while a provider run is in flight; `agent.run.cancel_requested`
+// (e.g. Slack's native stop button) aborts the controller. Streaming providers
+// honor the signal end-to-end (claude-code aborts the SDK run); accumulate-mode
+// runs that ignore it have their reply discarded on completion instead.
+const activeRunAborts = new Map<string, AbortController>();
+
+/**
+ * Abort the in-flight agent run and stream for a chat, if any.
+ * Exported for tests.
+ */
+export async function cancelActiveAgentRun(instanceId: string, chatId: string, reason: string): Promise<boolean> {
+  const key = `${instanceId}:${chatId}`;
+  const controller = activeRunAborts.get(key);
+  const sender = activeStreams.get(key);
+
+  if (!controller && !sender) {
+    log.info('Agent run cancel requested but nothing in flight', { instanceId, chatId, reason });
+    return false;
+  }
+
+  log.info('Cancelling in-flight agent run', {
+    instanceId,
+    chatId,
+    reason,
+    hasAbortController: !!controller,
+    hasActiveStream: !!sender,
+  });
+
+  controller?.abort();
+  if (sender) {
+    try {
+      await sender.abort();
+    } catch (err) {
+      log.warn('Stream sender abort failed during run cancellation', { instanceId, chatId, error: String(err) });
+    }
+  }
+  return true;
+}
+
 /** Route a single StreamDelta to the appropriate StreamSender method. */
 async function routeStreamDelta(sender: StreamSender, delta: StreamDelta): Promise<void> {
   switch (delta.phase) {
@@ -2431,6 +2472,12 @@ async function dispatchViaStreamingProvider(
   const streamKey = `${instance.id}:${chatId}`;
   activeStreams.set(streamKey, sender);
 
+  // Cancellable run (#914): agent.run.cancel_requested aborts this controller,
+  // which providers propagate into the underlying SDK run.
+  const cancelController = new AbortController();
+  activeRunAborts.set(streamKey, cancelController);
+  trigger.abortSignal = cancelController.signal;
+
   const streamDispatchStart = Date.now();
   try {
     const generator = resolved.provider.triggerStream(trigger);
@@ -2452,8 +2499,17 @@ async function dispatchViaStreamingProvider(
       });
     }
 
-    return streamResult;
+    // A cancelled run counts as handled even when the stream ended in an
+    // error delta — falling back would re-run the turn the user stopped.
+    return cancelController.signal.aborted ? true : streamResult;
   } catch (err) {
+    // A user-requested cancellation surfaces as an abort error from the
+    // provider — that's a handled outcome (#914), never a fallback: falling
+    // back to the accumulate path would re-run the turn the user just stopped.
+    if (cancelController.signal.aborted) {
+      log.info('Streaming dispatch cancelled by user', { instanceId: instance.id, chatId, traceId });
+      return true;
+    }
     log.error('Streaming dispatch failed, falling back', {
       instanceId: instance.id,
       chatId,
@@ -2468,6 +2524,7 @@ async function dispatchViaStreamingProvider(
     return false;
   } finally {
     activeStreams.delete(streamKey);
+    activeRunAborts.delete(streamKey);
   }
 }
 
@@ -2799,6 +2856,63 @@ async function dispatchViaTurnBasedProvider(
 }
 
 /**
+ * Why an accumulate-mode reply is withheld instead of sent.
+ * Precedence mirrors the original inline checks: a handoff announcement
+ * always wins over discard reasons.
+ */
+function resolveReplySuppression(flags: {
+  handoffTriggered: boolean;
+  errorHandoffDone: boolean;
+  supersededByNewerInbound: boolean;
+  runCancelled: boolean;
+}): 'handoff_triggered' | 'error_handoff' | 'superseded_by_newer_inbound' | 'run_cancelled_by_user' | null {
+  if (flags.handoffTriggered) return 'handoff_triggered';
+  if (flags.errorHandoffDone) return 'error_handoff';
+  if (flags.supersededByNewerInbound) return 'superseded_by_newer_inbound';
+  if (flags.runCancelled) return 'run_cancelled_by_user';
+  return null;
+}
+
+/**
+ * Run provider.trigger() with cancellation registered in activeRunAborts (#914).
+ *
+ * Returns null when the run was cancelled AND the provider surfaced the abort
+ * as a throw — the turn is over, nothing to send. Otherwise returns the result
+ * plus whether a cancel arrived mid-run (callers discard the reply then).
+ */
+async function triggerProviderWithCancellation(
+  provider: IAgentProvider,
+  trigger: AgentTrigger,
+  spanAttributes: Parameters<typeof withLifecycleSpan>[1],
+  instanceId: string,
+  chatId: string,
+  traceId: string,
+): Promise<{ result: Awaited<ReturnType<IAgentProvider['trigger']>>; cancelled: boolean } | null> {
+  const cancelController = new AbortController();
+  const runKey = `${instanceId}:${chatId}`;
+  activeRunAborts.set(runKey, cancelController);
+
+  try {
+    const result = await withLifecycleSpan('omni.dispatch_to_agno', spanAttributes, () =>
+      provider.trigger({
+        ...trigger,
+        abortSignal: cancelController.signal,
+        traceContext: activeProviderTraceContext() ?? trigger.traceContext,
+      }),
+    );
+    return { result, cancelled: cancelController.signal.aborted };
+  } catch (err) {
+    if (cancelController.signal.aborted) {
+      log.info('Agent run cancelled by user during dispatch', { instanceId, chatId, traceId });
+      return null;
+    }
+    throw err;
+  } finally {
+    activeRunAborts.delete(runKey);
+  }
+}
+
+/**
  * Try IAgentProvider dispatch first, return true if handled.
  * Falls back to legacy agentRunner.run() if provider not resolved.
  */
@@ -2943,15 +3057,23 @@ async function dispatchViaProvider(
       // ── Standard (round-trip / fire-and-forget) dispatch ──
       const correlationId = messages[0]?.metadata.correlationId;
       const dispatchStart = Date.now();
-      const result = await withLifecycleSpan(
-        'omni.dispatch_to_agno',
+
+      // Cancellable run (#914): providers that honor abortSignal stop mid-run;
+      // for the rest, the cancelled flag discards the reply after completion.
+      const triggerOutcome = await triggerProviderWithCancellation(
+        provider,
+        trigger,
         buildLifecycleSpanAttributes({
           ...lifecycleBase,
           stage: 'dispatch_to_agno',
           extra: { trigger_type: triggerType, provider_id: provider.id, provider_schema: provider.schema },
         }),
-        () => provider.trigger({ ...trigger, traceContext: activeProviderTraceContext() ?? trigger.traceContext }),
+        instance.id,
+        chatId,
+        traceId,
       );
+      if (!triggerOutcome) return true;
+      const { result, cancelled: runCancelled } = triggerOutcome;
       const dispatchDurationMs = Date.now() - dispatchStart;
 
       // Sentry metrics: agent dispatch count and latency
@@ -2992,7 +3114,14 @@ async function dispatchViaProvider(
         messages[0]?.metadata.trustedTenantId,
       );
 
-      if (result && result.parts.length > 0 && !handoffTriggered && !errorHandoffDone && !supersededByNewerInbound) {
+      const suppressReason = resolveReplySuppression({
+        handoffTriggered,
+        errorHandoffDone,
+        supersededByNewerInbound,
+        runCancelled,
+      });
+
+      if (result && result.parts.length > 0 && !suppressReason) {
         const selfChat = isSelfChat(chatId, instance.ownerIdentifier);
         const rawParts = selfChat ? result.parts.map((p) => `${BOT_PREFIX}${p}`) : result.parts;
         // Apply before_message_write hooks to each response part before sending
@@ -3032,13 +3161,9 @@ async function dispatchViaProvider(
 
         // T10: Agent chaining — forward response to chained instance if configured
         await forwardToChainedInstance(instance, parts, correlationId, messages);
-      } else if (handoffTriggered) {
-        log.info('Agent response suppressed — handoff triggered during run', {
-          instanceId: instance.id,
-          chatId,
-        });
-      } else if (supersededByNewerInbound) {
-        log.info('Agent response discarded — superseded by newer inbound', {
+      } else if (suppressReason) {
+        log.info('Agent response suppressed', {
+          reason: suppressReason,
           instanceId: instance.id,
           chatId,
           parts: result?.parts.length ?? 0,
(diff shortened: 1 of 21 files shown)


packages/channel-slack/src/handlers/typing.ts
```diff
@@ -31,12 +74,58 @@ export async function setSlackThreadStatus(params: {
...
+      await client.apiCall('agents.sessions.setStatus', {
+        channel_id: channelId,
+        thread_ts: threadTs,
+        // The new API takes a lifecycle enum, not a freeform string:
+        // non-empty legacy status → 'processing', clear → 'active'.
+        status: clearing ? 'active' : 'processing',
+      });
+      return { delivered: true, method: 'agents.sessions.setStatus' };
+    } catch (err) {
+      const code = slackErrorCode(err);
```
vasconceloscezar, line 127: **Fallback to the legacy status API only fires for 3 error codes, so most real failures kill status entirely.**

`AGENT_API_UNAVAILABLE_ERRORS` is `{unknown_method, feature_disabled, method_deprecated}`. Any other Slack platform error — `missing_scope` (the manifest's `REQUIRED_BOT_SCOPES` was not extended for the Agent Sessions API), `not_allowed_token_type` (exactly the code the pre-existing test in this file simulates, and what a user-token `actingClient` under `authMode: 'user'` will return), `invalid_auth`, `method_not_supported_for_channel_type` — takes the `else` branch: it logs a warn and returns `{ delivered: false, method: 'agents.sessions.setStatus' }` **without** trying `assistant.threads.setStatus`.

Scenario: any existing Slack app still on `assistant_view` / without the new scope loses thread status completely after this PR, where it worked before. Worse, the client is not added to `agentApiUnavailable`, so every subsequent status update pays another failing round-trip and never falls back.

The `else` branch should also fall through to the legacy path (and mark the client) for any error, not just these three codes.

packages/channel-slack/src/handlers/typing.ts
```diff
@@ -31,12 +74,58 @@ export async function setSlackThreadStatus(params: {
...
+    threadTs,
+    clearing,
+    statusLength: status.length,
+    loadingMessageCount: loadingMessages?.length ?? 0,
+  };
 
+  // ── Agent Sessions API (preferred) ──
+  if (typeof client.apiCall === 'function' && !agentApiUnavailable.has(client)) {
+    try {
+      await client.apiCall('agents.sessions.setStatus', {
```
vasconceloscezar, line 115: **The agents.sessions path silently drops the caller's status text and `loading_messages`, while the API still reports them as delivered.**

This call sends only `status: 'processing' | 'active'`; `loadingMessages` and the freeform `status` (e.g. `"analisando contexto..."`) supplied through `POST /messages/send/presence` (`packages/api/src/routes/v2/messages.ts:2463`) are never forwarded. `SlackPlugin.sendPresenceStatus` (`packages/channel-slack/src/plugin.ts:610`) still echoes `status` and `loadingMessages` back, so the endpoint responds `delivered: true, status: "analisando contexto...", loadingMessages: [...]` while Slack shows a generic processing state.

Either forward the fields (Slack's `agents.sessions.setStatus` accepts a status / loading-message payload) or stop reporting them as applied.

packages/api/src/plugins/agent-dispatcher.ts
```diff
@@ -6098,7 +6223,35 @@ export async function setupAgentDispatcher(
...
+              instanceId: payload.instanceId,
+              chatId: payload.chatId,
+              error: String(error),
+            });
+          }
+        });
+      },
+      {
+        durable: 'agent-dispatcher-cancel',
+        queue: 'agent-dispatcher-cancel',
```
vasconceloscezar: **The cancel subscription uses a NATS queue group, but the abort state is per-process, so the stop button is a coin flip in prod.**

`queue: 'agent-dispatcher-cancel'` becomes `deliver_group` (`packages/core/src/events/nats/consumer.ts:72`), so exactly one replica receives each `agent.run.cancel_requested`. `activeRunAborts` / `activeStreams` are module-level maps in the replica that is actually running the turn. `deploy/helm/omni/values-prod.yaml:58` sets `replicaCount: 2` (plus HPA), so roughly half of all stop presses land on the wrong replica and only log "nothing in flight" — the run keeps burning tokens.

This needs broadcast delivery (every replica subscribes independently and no-ops when it has nothing in flight, which `cancelActiveAgentRun` already does). Note that `withIdempotency` on the same handler actively prevents that fix: with a fan-out subscription only the first replica to claim `event.id` would run the abort.

packages/api/src/plugins/agent-dispatcher.ts
```diff
@@ -6098,7 +6223,35 @@ export async function setupAgentDispatcher(
...
     );
 
-    log.info('Agent dispatcher initialized (message + reaction + reaction-removed + media triggers)');
+    // ========================================
+    // Subscribe to agent.run.cancel_requested (#914)
+    // ========================================
+    await eventBus.subscribe(
+      'agent.run.cancel_requested',
+      async (event) => {
+        // Idempotency guard (#411): a replayed cancel must not abort a NEW
```
vasconceloscezar: **This comment does not describe the race the guard actually covers.**

`withIdempotency` only dedupes replays of the *same* `event.id`; it does nothing for a first-delivery cancel that arrives late.

Scenario: user presses stop, then immediately sends a new message; the NATS hop plus the `processed_events` insert take long enough that the new turn has already registered its controller under the same `instanceId:chatId` key, and `cancelActiveAgentRun` aborts the new run instead.

A run-scoped token (traceId/runId in the payload, compared before aborting) is what actually prevents this.

vasconceloscezar (commented):
Code review of `f80f501a`. 8 findings — 6 anchored inline below, 2 more here because they land outside the diff.

---

**Cancelling deletes the partial answer already shown, and post-abort deltas hit a deleted message** — `packages/channel-slack/src/senders/stream.ts:247` (and `:210`)

`cancelActiveAgentRun` now calls `sender.abort()` *mid-stream* (previously abort only ran after the generator ended). In replace and `status_final` modes, `abort()` does `chat.delete({ ts: draftTs })` — so pressing Slack's stop button erases everything the agent had already streamed, which is the opposite of Slack's stop semantics (halt and keep).

Additionally, for any provider that ignores `abortSignal` the generator keeps yielding: `onContentDelta` still sees `draftTs` set and issues `chat.update` against a deleted timestamp on every delta (swallowed as warns), and `native-stream.ts` keeps calling `streamer.append` after `stop()`. The senders need an "aborted" flag that makes subsequent deltas no-ops.

Relatedly, `streamingMessageTs` is parsed, typed and tested in `handlers/agent-sessions.ts` but dropped at the `handleAgentSessionStopped` boundary — it is exactly the data needed to know which streams Slack already halted.

---

**Presence response now mislabels Slack failures as `typing_indicator`** — `packages/api/src/routes/v2/messages.ts:2490`

`SlackPlugin.sendPresenceStatus` returns `{ delivered: false, reason }` with no `method` for `not_connected` / `no_active_thread` (`plugin.ts:545`, `:557`). The route's `presenceResult?.method ?? 'typing_indicator'` therefore reports `method: 'typing_indicator'` for a channel whose `canSendTyping` is `false` — previously it reported `assistant.threads.setStatus`. Any client keying off `method` to decide whether a real typing indicator was used is now misled on exactly the failure paths.

packages/api/src/plugins/agent-dispatcher.ts
```diff
@@ -2798,6 +2855,63 @@ async function dispatchViaTurnBasedProvider(
...
+  provider: IAgentProvider,
+  trigger: AgentTrigger,
+  spanAttributes: Parameters<typeof withLifecycleSpan>[1],
+  instanceId: string,
+  chatId: string,
+  traceId: string,
+): Promise<{ result: Awaited<ReturnType<IAgentProvider['trigger']>>; cancelled: boolean } | null> {
+  const cancelController = new AbortController();
+  const runKey = `${instanceId}:${chatId}`;
+  activeRunAborts.set(runKey, cancelController);
```
vasconceloscezar: **`activeRunAborts` key collision cancels the wrong run and orphans the other.**

The key is `${instanceId}:${chatId}`, and for Slack `chatId` is the *channel* id (`packages/channel-slack/src/handlers/messages.ts:208` → `chatId: meta.channelId`), not the thread. Two threads in the same channel are two concurrent runs: the streaming guard at line 2151 pushes the second one to the accumulate path, which then `activeRunAborts.set(runKey, …)` over the streaming run's controller.

Result: (a) pressing stop in thread A aborts thread B's run; (b) whichever run finishes first deletes the shared key in its `finally`, leaving the other run permanently uncancellable.

Keying by thread (the event already carries `threadId`) would fix both.

packages/channel-slack/src/manifest.ts
```diff
@@ -134,6 +147,13 @@ export function buildSlackManifest(options?: {
         display_name: displayName,
         always_online: true,
       },
+      // Agent messaging experience (#914). `assistant_view` is deprecated
+      // (removed February 2027) and new Slack apps can only use `agent_view`.
+      // Slack caps agent_description at 300 chars.
+      agent_view: {
```
vasconceloscezar: **`agent_view` is emitted unconditionally for an irreversible, app-wide migration.**

Every `buildSlackManifest()` call now includes `features.agent_view`, and this PR's own docs (`docs/channels/slack.md:9-13`) state the `assistant_view` → `agent_view` switch is one-way per app. An operator regenerating a manifest to add a slash command to an existing app silently and permanently migrates that app's experience. Every other optional behavior here is opt-in (`includeUserScopes`); this should be too.

Minor, next line: `.slice(0, 300)` truncates mid-word / mid-grapheme rather than trimming at a boundary.

packages/channel-slack/src/handlers/typing.ts
vasconceloscezar: Fixed in 2aa694bd. The `else` branch no longer returns early — every `agents.sessions.setStatus` failure now falls through to the legacy `assistant.threads.setStatus` bridge, so any app still on `assistant_view` (or missing the new scope) keeps its thread status working. On top of that, `missing_scope` and `not_allowed_token_type` were added to `AGENT_API_UNAVAILABLE_ERRORS`: those can never succeed on retry for that client, so they memoize it into `agentApiUnavailable` and subsequent calls skip the failing round-trip entirely. Transient/unknown errors fall back for that call but still retry the new API next time. New cases in `typing.test.ts` cover both behaviors (fall-through on arbitrary errors, memoization on auth-shaped ones).

packages/channel-slack/src/handlers/typing.ts
vasconceloscezar: Addressed in 2aa694bd via the second option ("stop reporting them as applied"). Forwarding isn't possible: checked against Slack's API reference — `agents.sessions.setStatus` takes only the session lifecycle enum and has no freeform status or loading-messages fields. So `sendPresenceStatus` now computes `appliedStatus`/`appliedLoadingMessages`: when the Agent Sessions API delivered, the response echoes the session enum (`processing`/`active`) and omits `loadingMessages`; the caller's freeform text and loading messages are only echoed back when the legacy `assistant.threads.setStatus` path actually delivered them (failed calls still echo the attempted values, flagged `delivered: false`, for diagnostics). The `POST /messages/send/presence` response no longer claims `"analisando contexto..."` was applied when Slack showed the generic processing state.

packages/api/src/plugins/agent-dispatcher.ts
vasconceloscezar: Fixed in 2aa694bd exactly as suggested: the subscription is now broadcast. The `queue`/`deliver_group` and the `durable` are gone, and the `withIdempotency` wrapper was removed too — as you noted, it would elect a single winner per `event.id` and defeat the fan-out. Each process now gets its own ephemeral consumer with `startFrom: 'new'` (no replay on restart), every replica sees the cancel, and the ones with nothing in flight no-op inside `cancelActiveAgentRun`. The replay protection that idempotency used to provide is replaced by the `requestedAt` guard (see the other thread on this file): a redelivered cancel compares the stop press time against each run's `startedAt` and cannot abort a newer run. New tests in `agent-dispatcher.test.ts` stage in-flight runs and assert the routing.

packages/api/src/plugins/agent-dispatcher.ts
vasconceloscezar: Fixed in 2aa694bd with the run-scoped token you describe, implemented as a timestamp rather than a runId — Slack's `agent_session_stopped` event doesn't carry our traceId/runId, but it does carry `event_ts`, which is exactly the instant of the stop press. `AgentRunCancelRequestedPayload` gains `requestedAt` (the parsed `event_ts` in unix ms, from `plugin.ts`), each `activeRunAborts` entry now records `startedAt`, and `cancelActiveAgentRun` refuses to abort a run whose `startedAt` is after `requestedAt`. So in your scenario — stop pressed, new message sent, cancel arrives late — the new turn registered with a later `startedAt` and the cancel is logged and ignored. The misleading `withIdempotency` comment (and the wrapper itself) is gone with the broadcast-subscription change. Covered by a new test asserting a late cancel leaves the newer run alone.

packages/api/src/plugins/agent-dispatcher.ts
vasconceloscezar: Fixed in 2aa694bd by keying on the thread, as suggested. `activeRunAborts` is now keyed `runAbortKey(instanceId, chatId, threadId)` and stores `{ controller, startedAt, sender }`; both registration sites — the streaming path and `triggerProviderWithCancellation` (accumulate) — use the thread-scoped key. That fixes both failure modes: a stop in thread A can no longer abort thread B's run, and each run's `finally` deletes only its own entry, so neither run orphans the other. `cancelActiveAgentRun` looks up the exact thread key first, then falls back to the threadless key for runs started from a top-level mention/DM that registered without a `thread_ts`. `__test__` now exposes `activeRunAborts`/`runAbortKey`, and new tests in `agent-dispatcher.test.ts` cover two concurrent threads in one channel.

packages/channel-slack/src/manifest.ts
vasconceloscezar: Addressed in 2aa694bd, with one deliberate deviation. `buildSlackManifest` now takes an `agentView` option so the block is conditional, but the default is `true` rather than opt-in. Reasoning: this generator never emitted `assistant_view`, so any app created from its manifests is already on `agent_view` (or was created after Slack made `agent_view` the only option for new apps) — defaulting off would silently strip the agent experience from the generator's own apps to protect apps it didn't create. The dangerous case you describe (regenerating a manifest for a hand-migrated `assistant_view` app) is covered by the explicit `agentView: false` opt-out, documented both on the option and in `docs/channels/slack.md` with the one-way-migration warning. Happy to flip the default if you still prefer opt-in.

The truncation nit is fixed too: `.slice(0, 300)` is replaced by `truncateAtWordBoundary`, which cuts at the last space and, when the text has no spaces, hard-cuts without splitting a surrogate pair.

Issue #1105, opened 2026-09-11
chore(channel-whatsapp): bump vendored baileys 7.0.0-rc10 → rc14 (brings the protocol-message spoofing guard)


vasconceloscezar:
## Summary

`packages/channel-whatsapp` vendors **baileys 7.0.0-rc10** (`vendor/baileys-v7.0.0-rc10.tgz`); upstream latest is **7.0.0-rc14**. Four RCs behind on the library that carries every WhatsApp instance.

## What rc11–rc14 change (diffed locally against our vendored copy)

`lib/Utils/process-message.js` (+33 lines) and `lib/Utils/decode-wa-message.js` both changed. The substantive one is a **spoofing guard on protocol messages**, mirroring whatsmeow's `handleProtocolMessage`: self-only types are dropped when `!fromMe` —

- `HISTORY_SYNC_NOTIFICATION`, `APP_STATE_SYNC_KEY_SHARE`, `LID_MIGRATION_MAPPING_SYNC`, `PEER_DATA_OPERATION_REQUEST_RESPONSE_MESSAGE`

— because a third party could otherwise forge them to manipulate local state (history sync, app-state keys, LID mapping). Cross-user types (`REVOKE`, `MESSAGE_EDIT`, `EPHEMERAL_SETTING`, `GROUP_MEMBER_LABEL_CHANGE`) are explicitly kept. That is a security fix we do not have today.

## What it does NOT fix

Checked explicitly because of #1061: rc14 still has **zero handling for `secretEncryptedMessage`** — the field exists in the protobuf (`WAProto.proto:2134`, field 82, with `SecretEncType.MESSAGE_EDIT = 2`) and is referenced nowhere under `lib/`. Encrypted edits stay opaque on rc14 exactly as on rc10, so this upgrade is not a path to closing #1061.

## Ask

Bump the vendored tarball to 7.0.0-rc14, run the channel suite (394 tests) plus a live smoke on one instance (inbound/outbound text, media, reaction, delete, group vs DM, LID chat), and re-vendor. Worth pairing with a note in the package README on how the vendored copy is refreshed, since a pinned tarball has no update signal — this one drifted four RCs silently.

PR #1106, opened 2026-09-11
fix(whatsapp): land message edits — upsert protocol path + msgSecret decryption (#1061)

### LINKED ISSUE #1061: bug(channels): message edits journal as contentType unknown with empty text — edited content lost
## Summary

Editing a WhatsApp message journals a `message.received` with `contentType: "unknown"` and **empty `textContent`** — the edit is not recognized as an edit and the new text is lost. The messages table has edit support (`editCount`, `originalText`, `editHistory` fields exist), but the event journal treats the edit as an unknown payload.

## Repro (v2.260910.4, whatsapp-baileys)

Edit a previously sent message in a monitored chat → journal gains `{eventType: message.received, contentType: "unknown", textContent: ""}` (verified on both observing instances). No `message.edited`-style event; no edited text anywhere in the event.

## Expected

Either a dedicated edit event (`message.edited` with new text + target external id — consistent with the reaction.* direction taken in #1045) or, minimally, `contentType: "edit"` with the new text in `textContent` and the target message id in payload.

## Related

#1041 (same family: non-message kinds under message.received), #1033/#1059 (taxonomy for reactions).

vasconceloscezar:
## Problem

An edit made on **another device of the same account** syncs as a `MESSAGE_EDIT` protocol message on `messages.upsert` and never produces a `messages.update`. The upsert branch swallowed it on purpose — waiting for an update that never arrives — so that edit was lost entirely: no journal row, `editCount` stayed 0, the pre-edit text stayed.

The upsert already extracted the new text into `content.editedText` and discarded it, so the data was there the whole time.

## Change

- `handleSpecialMessage` emits the edit when the upsert already carries the new text; it still defers to `messages.update` when it does not (protocol message without `editedMessage`).
- A bounded seen-set keyed by `(targetMessageId, newText)` de-duplicates: a third-party edit reaches us on **both** paths, and only the first arrival emits. Keyed by text so a genuine second edit of the same message still passes; capped at 500 entries so a long-lived socket cannot grow it without end.

## Tests

`packages/channel-whatsapp/src/__tests__/message-edit-self-echo.test.ts` — 4 cases:
1. edit arriving only as an upsert protocol message is emitted;
2. same edit arriving on both paths emits once;
3. a genuine second edit of the same message still emits;
4. a textless upsert edit still defers to `messages.update`.

Channel suite green: **394 pass / 0 fail** (30 files), including the existing `message-edit-update.test.ts`.

## Scope — read before closing anything with this

This does **not** fix #1061's reproduction. Capturing the raw payload of a failing edit on v2.260910.18 showed it arrives as an encrypted secret envelope, not as a protocol message:

```jsonc
"secretEncryptedMessage": {
  "encIv": "…", "encPayload": "…",
  "secretEncType": "MESSAGE_EDIT",
  "targetMessageKey": { "id": "<original-msg-id>", … }
}
```

The new text is encrypted under the original message's `messageContextInfo.messageSecret` (same family as poll votes / event responses). Baileys — including latest rc14, checked — ships **no handler for `secretEncryptedMessage`**; the field exists in the protobuf (`WAProto.proto:2134`, `SecretEncType.MESSAGE_EDIT = 2`) and is referenced nowhere in `lib/`. Deriving the key by analogy with `decryptPollVote` failed across 288 combinations of info string × JID × AAD shape, so it needs the actual scheme rather than guesswork. Full analysis and payload in #1061.

So: this PR closes a real, separate gap in the plaintext-protocol path. #1061 should stay open for the encrypted case (and is likely an upstream Baileys feature).

Related: #1105 (bump vendored baileys rc10 → rc14 — brings a protocol-message spoofing guard; does not help here).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WhatsApp message edit handling to prevent duplicate edit events when updates arrive through multiple synchronization paths.
  * Preserved separate notifications for genuine subsequent edits to the same message.
  * Ensured edits lacking updated text continue through the appropriate processing path.

### DIFF
--- a/apps/ui/dist-node/vite.config.js
+++ b/apps/ui/dist-node/vite.config.js
@@ -13,7 +13,8 @@ export default defineConfig({
         port: 5173,
         proxy: {
             '/api': {
-                target: 'http://localhost:8882',
+                // Override when the API runs on a non-default port (e.g. 8882 taken).
+                target: process.env.OMNI_API_PROXY_TARGET || 'http://localhost:8882',
                 changeOrigin: true,
                 configure: (proxy) => {
                     // Fix ERR_CONTENT_DECODING_FAILED by removing accept-encoding
--- a/packages/channel-whatsapp/src/__tests__/message-edit-encrypted.test.ts
+++ b/packages/channel-whatsapp/src/__tests__/message-edit-encrypted.test.ts
@@ -1,0 +1,111 @@
+/**
+ * Encrypted edits end to end (#1061).
+ *
+ * WhatsApp delivers an edit as a `secretEncryptedMessage` whose payload is encrypted
+ * under the ORIGINAL message's `messageSecret`. This exercises the whole inbound path:
+ * the original message lands (its secret is remembered), the encrypted edit lands, and
+ * the decrypted new text reaches `handleMessageEdited`.
+ *
+ * The ciphertext is a REAL captured edit; the expected plaintext is its edited text.
+ */
+
+import { describe, expect, it, mock } from 'bun:test';
+import { EventEmitter } from 'node:events';
+import { setupMessageHandlers } from '../handlers/messages';
+import { WhatsAppPlugin } from '../plugin';
+
+const GROUP = '120363406854035156@g.us';
+const SENDER_LID = '217046273028329@lid';
+const ORIG_ID = '3B335662F99E0B247ABE';
+const SECRET = Buffer.from('MGULupwDcKGlKJDlJlZ+DlAyld9U+VmxvuJI1XNrgf8=', 'base64');
+const ENC_IV = Buffer.from('L50Zn/v7vdNibfeL', 'base64');
+const ENC_PAYLOAD = Buffer.from(
+  'qDEc9XltNsL19474MjwjA027q7mtHgRhrsYUvc29lMdrsi0NmgZ7HiItLsW8I1DJ2az/dc/al/y1Yfxkt5y88NzT9cs1OwrcgCAV/XVHANJZ/umo3M8OvKlX7FARPVjUZRvKZm+fBOs9KmB+MSfLj+lVplqjBKmh5GYBJ3zdSeN4Ht/ZgoBpaMme307JGNhCn0xGTJW2fziOkBQhlHz6GdaxIhw=',
+  'base64',
+);
+
+function harness() {
+  const ev = new EventEmitter();
+  const sock = { ev, presenceSubscribe: mock(async () => {}) } as unknown as Parameters<typeof setupMessageHandlers>[0];
+  const plugin = new WhatsAppPlugin();
+  (plugin as unknown as { config: unknown }).config = { apiBaseUrl: 'http://localhost:8882' };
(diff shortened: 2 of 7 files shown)


vasconceloscezar:
## Update: this PR now also decrypts the encrypted edits — #1061 is fully covered

The scope note above said the encrypted case needed the actual scheme rather than guesswork. Found it, implemented it, and validated it against live WhatsApp traffic.

### Root cause

WhatsApp moved message edits to the **msgSecret envelope**. The edit arrives as `secretEncryptedMessage { encIv, encPayload, secretEncType: MESSAGE_EDIT, targetMessageKey }`, with the new content encrypted under the **original message's** `messageContextInfo.messageSecret`. Baileys carries the protobuf (`SecretEncType.MESSAGE_EDIT = 2`) but ships **no decryptor** — verified through latest rc14 — so the envelope reached us opaque, journaled as `contentType: unknown`, and the edit was lost. That is why both previous fixes, which target `protocolMessage` / `messages.update`, could not land it.

### Scheme (matches whatsmeow's `generateMsgSecretKey`, msgsecret.go)

```
useCase = origMsgId ‖ origSenderJid ‖ modificationSenderJid ‖ "Message Edit"
key     = HKDF-SHA256(ikm = origMessageSecret, salt = ∅, info = useCase, len = 32)
text    = AES-256-GCM(key, encIv, encPayload)        // no AAD — only Poll Vote
                                                     // and Event Response use one
```

Both JIDs in non-AD form (`user@server`). The plaintext is a full `proto.Message` whose `protocolMessage.editedMessage` holds the new content. Earlier attempts failed on three details worth recording: an extra `0x01` byte in the useCase (not part of this scheme), bare user numbers instead of full JIDs, and passing an AAD where edits use none.

### Implementation

- `utils/msg-secret.ts` — derivation + decryption, returning `null` on any failure (never throws).
- Inbound messages remember their own `messageSecret` in a bounded in-process map (2000 entries, insertion-ordered), because the edit envelope never repeats it. This mirrors the durable `MsgSecrets` store whatsmeow keeps; a miss (secret never seen, process restart) degrades exactly to the previous behaviour.
- A `secretEncryptedMessage` extractor surfaces the envelope; decryption happens in the handler, which can reach the stored secret.
- `handleMessageEdited` now takes the editing sender's JID, so `editHistory[].by` records the author instead of the chat.

### Validation

**Live, on a real instance:** edited a message in a WhatsApp group; the journal took the edit and the message row became

```json
{ "editCount": 1,
  "textContent":  "opa, será mesmo q sabe quando eu editei?",
  "originalText": "opa",
  "editHistory": [{ "at": "…", "text": "opa, será mesmo q sabe quando eu editei?" }],
  "status": "edited" }
```

**Tests:** a real captured ciphertext is the fixture — unit tests for the derivation (plus wrong-secret and wrong-sender-form paths returning null) and an end-to-end test through the inbound handler (original message seeds the secret → encrypted edit → decrypted text reaches `handleMessageEdited`), plus the no-secret path staying silent. Channel suite: **429 pass / 0 fail**.

Closes #1061.

